PR #8, opened 2021-01-21
burn/mint liquidity argument consistency

stanieltron:
modified burn_liquidity to accept first_asset_amount as parameter instead of liquidity_asset_amount


gleb-urvanov (approved):
gj

PR #10, opened 2021-02-08
Feature/test fix and cleanup

stanieltron:
u256 math for all fns
test completetion, rehaul and tested with large values


gleb-urvanov (changes requested):
`cargo test` is failing for the pallet

pallets/xyk/src/lib.rs
```diff
@@ -1,36 +1,28 @@
...
+use sp_runtime::traits::{BlakeTwo256, Hash, One, SaturatedConversion, Zero};
 #[cfg(test)]
 mod mock;
-
 #[cfg(test)]
 mod tests;
-
 pub trait Trait: assets::Trait {
     // TODO: Add other types and constants required configure this module.
+    // type Randomness: Randomness<Self::Hash>;
```
gleb-urvanov, line 21: why it is back?

PR #12, opened 2021-02-10
Feature/pool address

stanieltron:
zero amount ensures
vault acc amount and zero amount check

### DIFF
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -3,15 +3,15 @@
 use codec::{Decode, Encode};
 use frame_support::{
     decl_error, decl_event, decl_module, decl_storage, dispatch::DispatchResult, ensure,
-    weights::Pays, StorageMap, sp_runtime::ModuleId,
+    sp_runtime::ModuleId, weights::Pays, StorageMap,
 };
 use frame_system::ensure_signed;
 use pallet_assets as assets;
 use sp_core::U256;
 // TODO documentation!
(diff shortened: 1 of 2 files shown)


pallets/xyk/src/lib.rs
```diff
@@ -21,6 +22,7 @@ pub trait Trait: assets::Trait {
     type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
 }
 
+const PALLET_ID: ModuleId = ModuleId(*b"Treasury");
```
gleb-urvanov, line 25: please use a hex string as an id

pallets/xyk/src/lib.rs
```diff
@@ -431,15 +451,12 @@ decl_module! {
     }
 }
 
-
 impl<T: Trait> Module<T> {
```
gleb-urvanov, line 454: set formatter to ignore empty lines, we should have them to visually split the code

pallets/xyk/src/lib.rs
```diff
@@ -33,19 +35,20 @@ decl_error! {
...
     {
         //TODO add trading events
         SomethingStored(u32, AccountId),
     }
 );
 
-
 // XYK exchange pallet storage.
 decl_storage! {
     trait Store for Module<T: Trait> as XykStorage {
```
gleb-urvanov, line 54: Remove VaultId from storage since it is no longer used

gleb-urvanov (changes requested):

PR #16, opened 2021-03-07
disabled asset issue functionality for user

stanieltron:
-asset issue fn converted to sudo function
-added recipient argument
-modified usage of this fn in other pallets

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -1,14 +1,18 @@
 // Copyright (C) 2020 Mangata team
 
-use sp_core::{Pair, Public, sr25519};
 use mangata_runtime::{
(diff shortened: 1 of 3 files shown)


gleb-urvanov:
@stanieltron @iStrike7 
Are changes introduced in this PR reflected in e2e tests? 
So the assets for the e2e are created as sudo, or as a user, or are not created at all?

iStrike7:
> @stanieltron @iStrike7
> Are changes introduced in this PR reflected in e2e tests?
> So the assets for the e2e are created as sudo, or as a user, or are not created at all?

Assets for e2e are created "as a user". We may be able to change this pretty easily (to using sudo).

PR #19, opened 2021-03-28
All working with bridge tests

iStrike7:
All hard coded values such as bridgedAssetId for MNG and ETH and the hardcoded appId for ERC20 and ETH apps, may need to be changed depending on the deployment.


primitives/core/src/registry.rs
```diff
@@ -20,11 +20,11 @@ struct Entry {
 static APP_REGISTRY: &[Entry] = &[
 	Entry {
 		app: App::ETH,
-		id: [0xdd, 0x51, 0x4b, 0xaa, 0x31, 0x7b, 0xf0, 0x95, 0xdd, 0xba, 0x2c, 0x0a, 0x84, 0x77, 0x65, 0xfe, 0xb3, 0x89, 0xc6, 0xa0],
+		id: [252, 151, 166, 25, 125, 201, 11, 239, 107, 190, 253, 103, 39, 66, 237, 117, 233, 118, 133, 83],
```
gleb-urvanov, line 23: Why app addresses are changed?

pallets/erc20-app/src/lib.rs
```diff
@@ -110,6 +112,10 @@ impl<T: Trait> Module<T> {
 			return Err(Error::<T>::InvalidAssetId.into())
 		}
 
+		if ( T::AccountId::default() == payload.recipient_addr ) {
```
gleb-urvanov, line 115: why don't allow transfer of erc20 to default address?

node/src/chain_spec.rs
```diff
@@ -72,6 +73,20 @@ pub fn development_config() -> Result<ChainSpec, String> {
 				"0xec00ad0ec6eeb271a9689888f644d9262016a26a25314ff4ff5d756404c44112"
 					.parse()
 					.unwrap(),
+				// SnowBridged Assets
+				vec![
```
gleb-urvanov, line 77: I can see the config for `bridged_assets` is defined, but the corresponding `asset` is not created. Am I missing the config for the assets pallet?

node/src/chain_spec.rs
```diff
@@ -72,6 +73,20 @@ pub fn development_config() -> Result<ChainSpec, String> {
...
+						(
+							"Mangata".as_bytes().to_vec(), "MNG".as_bytes().to_vec(), "Mangata Asset".as_bytes().to_vec(),
+							18u32, 0u32, H160::from_slice(&hex!["F8F7758FbcEfd546eAEff7dE24AFf666B6228e73"][..]),
```
gleb-urvanov, line 80: Is it mangata address for the local testnet or for Kovan?

node/src/chain_spec.rs
```diff
@@ -72,6 +73,20 @@ pub fn development_config() -> Result<ChainSpec, String> {
...
+							1_000_000u128, get_account_id_from_seed::<sr25519::Public>("Alice")
```
gleb-urvanov, line 81: Let there be 100 million Mangata tokens.
Shouldn't the total supply be in the smallest denomination?

gleb-urvanov (changes requested):

node/src/chain_spec.rs
iStrike7: Do you mean "u128 -> u32"?
I used u128 because u128 is what is set as Balance in runtime/lib.rs

node/src/chain_spec.rs
iStrike7: All addresses are for local testnet.

node/src/chain_spec.rs
iStrike7: The asset is being initialized in add_extra_genesis in pallests/bridged_asset.
This is because the bridged_asset pallet is tightly-coupled ("pub trait Trait: assets::Trait") to the assets pallet, it seemed natural to have the bridged_assets pallet call asset_issue from the assets pallet.

pallets/erc20-app/src/lib.rs
iStrike7: Default address corresponds to "0x00..00".
The bridge contract (on ethereum) locks and sends in the same step. So a dummy call must inevitable be made when initializing a to token on ethereum for the bridge the way we are for MNG.
To avoid this dummy call from actually doing anything, I call the lock and send function on ethereum using "subNullRecipient", and then these ignore calls in substrate.

primitives/core/src/registry.rs
iStrike7: These are testnet addresses. All addresses are testnet addresses.

gleb-urvanov: I suppose you mean the local testnet. But in this case, the node will not be working with Kovan.
Are these values parameterized using environment?

pallets/erc20-app/src/lib.rs
gleb-urvanov: But we can do the trick much simpler- instead of calling lock and send function on the contract, we can just transfer tokens to the contract using `transfer` function of token contract. As the result, the tokens will be locked on the contract address, but no bridge-related events are triggred

node/src/chain_spec.rs
gleb-urvanov: Here I mean the amount is set not by specifying the number of whole tokens, but by the amount of smallest, indivisible shares of it. So, if a token has 6 numbers after a decimal point and you want to have 1 token in total, you should specify 1_000_000 here, similar as you do in ethereum smartcontracts. Am I wrong?

node/src/chain_spec.rs
gleb-urvanov: Can you change the address for development chainspec to Kovan mangata address (when it will be deployed), and leave the local address for the local testnet?

node/src/chain_spec.rs
gleb-urvanov: Good approach, keep in mind the bond between the integer balance and smallest denomination balance

pallets/erc20-app/src/lib.rs
iStrike7: For the bridge to work the other way there must be an entry in the ERC20App contract's registry with sufficient amount in favor of the ERC20 token. So when the ERC20 tokens are being locked and sent from ethereum to substrate, the locking of the tokens must be added to the registry with "totalTokens[_tokenAddr] = totalTokens[_tokenAddr].add(_amount);"
This operation is only carried out in the function "sendERC20", which carries out both locking and sending.

primitives/core/src/registry.rs
iStrike7: Yes, I mean the local testnet. Yes, the node will not work with Kovan ootb. You will need to set variables as per your deployment.
No. Not yet.

node/src/chain_spec.rs
iStrike7: Oh right! I completely missed that!
Thank you for pointing that out.

node/src/chain_spec.rs
iStrike7: As per my understanding, all input, output, processing, computation and storage are done in the "smallest denomination balance", while the representation of these values is left to the ui/fe using the decimal information.
Am I right?
If yes, then I can just use the smallest denomination balance everywhere.

node/src/chain_spec.rs
iStrike7: I have configured everything to run with --dev as local doesn't seem to work.
When you start the node with the local config, it just keeps printing out:
INFO 💤 Idle (0 peers), best: #0 (0x5817…d9ad), finalized #0 (0x5817…d9ad), ⬇ 0 ⬆ 0

primitives/core/src/registry.rs
gleb-urvanov: Ok please revert the value to match public testnet's Kovan address

pallets/erc20-app/src/lib.rs
gleb-urvanov: I was used to think the balance of the contract address serves as a registry. I will doublecheck it and get back to this question

node/src/chain_spec.rs
gleb-urvanov: Yes, you are right, no operations on the node side should involve the decimal information.

node/src/chain_spec.rs
gleb-urvanov: I see. Please revert the parameters of the dev to match the public testnet since it is the configuration currently used there.

gleb-urvanov:
Can you please revert the cargo fmt for now? It makes impossible to review the changes.

primitives/core/src/registry.rs
iStrike7: What about the MNG contract address? That will require a deployment of the ERC20 MNG token on Kovan. Should I leave that as it is (setup for the local testnet)?

pallets/erc20-app/src/lib.rs
gleb-urvanov: See it now. Agree with your solution.

PR #21, opened 2021-04-05
Feature/mutable chainspecced ethereum app ids

iStrike7:


node/src/chain_spec.rs
```diff
@@ -73,6 +77,15 @@ pub fn development_config() -> Result<ChainSpec, String> {
 				"0xec00ad0ec6eeb271a9689888f644d9262016a26a25314ff4ff5d756404c44112"
 					.parse()
 					.unwrap(),
+				// Ethereum AppId for SnowBridged Assets
+				vec![
+						(
+							App::ETH, H160::from_slice(&hex!["Fc97A6197dc90bef6bbEFD672742Ed75E9768553"][..]).into()
```
gleb-urvanov, line 83: Are these addresses for local eth testnet of for kovan testnet?

gleb-urvanov:
Please resolve dead code warning at `primitives/core/build.rs:14`

primitives/core/src/lib.rs
```diff
@@ -10,10 +10,13 @@ use frame_support::dispatch::DispatchResult;
 
 use sp_std::vec::Vec;
 
+
+// use serde::{Serialize, Deserialize};
```
gleb-urvanov, line 14: remove unused code

node/Cargo.toml
```diff
@@ -81,5 +85,9 @@ xyk-runtime-api = { default-features = false, version = '2.0.0', path = '../pall
 hex-literal = { version = "0.3.1", default-features = false }
 
 [features]
-default = []
+default = ['std']
```
gleb-urvanov, line 88: why do we need `std` here?

node/src/chain_spec.rs
iStrike7: Thanks for catching that. Changed it to Kovan testnet values.

primitives/core/src/lib.rs
iStrike7: removed.

node/Cargo.toml
iStrike7: Since root project is a virtual workspace (doesn't specify a package) we can't run it with features.
To run runtime-benchmark feature we need to run it from /node.
Node builds xyk-rpc. Mangata-runtime doesn't.
xyk-rpc needs to be built after xyk-runtime-api which is built by mangata-runtime.
Thus we need: std = [
		"mangata-runtime/std",
		"xyk-rpc/std",
]
We add default = ['std'] simply to avoid having to type
--features "runtime-benchmarks std"

PR #24, opened 2021-04-08
fixes for 'cargo test' 

mateuszaaa:
- bunch of tests has been disabled due to incompatible Asset pallet (that has been replaced some time ago) - those where commented out with reference to proper trello ticket
- some of the test aligned with implementation
- now 2 tests are failing - could you @gleb-urvanov  take a look if thats expected behaviour ? i`m not sure if i need to fix those or just mark as ignored  for now?

```
failures:

---- tests::block_import_of_bad_extrinsic_root_fails stdout ----
on_initialize(1)
on_finalize(?)
Hash: given=49cd58a254ccf6abc4a023d9a22dcfc421e385527a250faec69f8ad0d8ed3e48, expected=465a1569d309039bdf84b0479d28064ea29e6584584dc7d788904bb14489c6f6
note: test did not panic as expected
---- tests::block_import_of_bad_state_root_fails stdout ----
on_initialize(1)
on_finalize(?)
Hash: given=0000000000000000000000000000000000000000000000000000000000000000, expected=465a1569d309039bdf84b0479d28064ea29e6584584dc7d788904bb14489c6f6
note: test did not panic as expected


```


pallets/erc20-app/src/mock.rs.orig
```diff
@@ -0,0 +1,140 @@
+// Mock runtime
```
gleb-urvanov, line 1: here and later- are these .orig files created intentionally?

gleb-urvanov:
block_import_of_bad_extrinsic_root_fails and block_import_of_bad_state_root_fails  tests fail is expected behavior, you can mark these as ignored but we are aiming to resolve the cause of failure later, so we should not forget to return the tests then.

pallets/erc20-app/src/mock.rs.orig
mateuszaaa: my bad - fixed already

gleb-urvanov:
block_period_minimum_enforced failure is expected. We should take a look at block timing later, it does not have a priority right now

mateuszaaa:
ok - i ignored it and created proper ticket

node/src/chain_spec.rs
```diff
@@ -91,19 +92,19 @@ pub fn development_config() -> Result<ChainSpec, String> {
                 // SnowBridged Assets
                 vec![
                     (
-                        "Mangata".as_bytes().to_vec(),
-                        "MNG".as_bytes().to_vec(),
-                        "Mangata Asset".as_bytes().to_vec(),
+                        "Mangata".to_string().into_bytes(),
```
gleb-urvanov, line 95: Does `.to_string().into_bytes()` produce the same output as `.as_bytes().to_vec()`? If so, why `.to_string().into_bytes()` is preferred?

node/src/chain_spec.rs
```diff
@@ -139,7 +140,8 @@ pub fn development_config() -> Result<ChainSpec, String> {
 }
 
 pub fn local_testnet_config() -> Result<ChainSpec, String> {
-    let wasm_binary = WASM_BINARY.ok_or("Development wasm binary not available".to_string())?;
+    let wasm_binary =
+        WASM_BINARY.ok_or_else(|| "Development wasm binary not available".to_string())?;
```
gleb-urvanov, line 144: What is the difference between `.ok_or(smth)` and `ok_or_else(|| smth)`? What is the rule to choose between two?

node/src/chain_spec.rs
```diff
@@ -175,19 +177,19 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
...
-                        "MNG".as_bytes().to_vec(),
-                        "Mangata Asset".as_bytes().to_vec(),
+                        "Mangata".to_string().into_bytes(),
+                        "MNG".to_string().into_bytes(),
+                        "Mangata Asset".to_string().into_bytes(),
                         18u32,
                         0u32,
                         H160::from_slice(&hex!["F8F7758FbcEfd546eAEff7dE24AFf666B6228e73"][..]),
-                        100_000_000__000_000_000_000_000_000u128,
+                        100_000_000_000_000_000_000_000_000u128,
```
gleb-urvanov, line 186: Can we suppress a warning here? The original constant `100_000_000__000_000_000_000_000_000u128` visually separates integer and decimal parts of the total amount

pallets/assets-info/src/lib.rs
```diff
@@ -71,7 +72,7 @@ decl_storage! {
                         name: name.clone(),
                         symbol: token.clone(),
                         description: description.clone(),
-                        decimals: decimals.clone(),
+                        decimals: decimals.to_owned(),
```
gleb-urvanov, line 75: Per my understanding, `.clone()` or `.to_owned()` is required because the value was borrowed. Why in case of `name`, `symbol` and `description` the right way to go is `.clone()` and for `decimals` it is `.to_owned()` ?

pallets/assets-info/src/tests.rs
```diff
@@ -20,7 +20,7 @@ fn set_info_and_retrieve_works_ok() {
             info.name.clone(),
             info.symbol.clone(),
             info.description.clone(),
-            info.decimals.clone(),
+            info.decimals,
```
gleb-urvanov, line 23: why decimals are treated differently compared to the rest fields of the info struct?

pallets/bridged-asset/src/tests.rs
```diff
@@ -1,224 +1,226 @@
...
+// where
+//     T: Into<U256> + Copy,
+// {
+//     let free = amount.into();
+//     let account = AccountData { free };
+//     Account::<MockRuntime>::insert(&asset_id, &account_id, &account);
+//     TotalIssuance::insert(&asset_id, free);
+// }
+//
+// #[test]
```
gleb-urvanov, line 28: these tests of bridged asset are no longer relevant since token functionality like mint, burn, and transfer is no longer implemented in bridged asset, it is no longer responsible for that functionality. Thus, these tests are to be removed.

pallets/erc20-app/src/tests.rs
```diff
@@ -1,62 +1,64 @@
...
+//
+// const RECIPIENT_ADDR_BYTES: [u8; 32] =
+//     hex!["8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48"];
+//
+// fn last_event() -> MockEvent {
+//     System::events().pop().expect("Event expected").event
+// }
+//
+// #[test]
+// fn mints_after_handling_ethereum_event() {
```
gleb-urvanov, line 25: FYI these should be relatively easy to fix, just by using proper Assets pallet (instead of outdated Asset)
Since Assets will be replaced soon, there are no need to do it now.

pallets/xyk/src/tests.rs
```diff
@@ -118,7 +120,7 @@ fn multi() {
             500000000000000000000000
         ); // amount of asset 1 in vault acc after creating pool
 
-        XykStorage::mint_liquidity(Origin::signed(2), 0, 1, 500000000000000000000000);
+        XykStorage::mint_liquidity(Origin::signed(2), 0, 1, 500000000000000000000000).unwrap();
```
gleb-urvanov, line 123: here and later- why `.unwrap()` is desired?

gleb-urvanov (approved):

PR #27, opened 2021-04-16
Feature/burn liquidity residual

stanieltron:
modified burn liquidity fn to accept liquidity asset amount and to calculate amounts through separate get_burn_amount fn, which is also accessible as rpc


gleb-urvanov (changes requested):
Undo styling alterations

PR #31, opened 2021-04-22
created chainspec for public testnet

gleb-urvanov:


omaksi:
1. What is the intended usage of this chainspec?
2. Also I'm a bit confused about the hardcoded localhost bootnode, what is the reason for it?

gleb-urvanov:
> What is the intended usage of this chainspec?  
It is for the public testnet, we should use it in the next testnet iteration. It replaces the testing keys we are currently using with private keys for our validator and relayer

> Also I'm a bit confused about the hardcoded localhost bootnode, what is the reason for it?  
My bad, should be removed

omaksi:
1. Which keys are being used now?
2. It seems the babe key is not //Alice but grandpa key is //Alice

gleb-urvanov:
> Which keys are being used now?

provided using secure channel
> It seems the babe key is not //Alice but grandpa key is //Alice

nice catch, I forgot about different scheme. Replaced.

PR #33, opened 2021-04-22
replace pallet-assets with tokens

mateuszaaa:


pallets/erc20-app/src/lib.rs
```diff
@@ -92,11 +96,23 @@ decl_module! {
                 return Err(Error::<T>::InvalidAssetId.into())
             }
 
-            // <asset::Module<T>>::do_burn(asset_id, &who, amount)?;
+
             let native_asset_id = <asset::Module<T>>::get_native_asset_id(asset_id);
-            let result = <assets::Module<T>>::assets_burn(&native_asset_id, &who, &amount.low_u128().saturated_into::<T::Balance>());
 
-            ensure!(result.is_ok(), Error::<T>::BurnFailure);
+            ensure!(T::Currency::can_slash(
```
iStrike7, line 102: I do think we can use slash instead of burn here, because are checking can_slash which checks for free balance availability.
Still maybe we should check with Gleb once.
Without can_slash check it might actually slash from reserved balance and might not even slash the entire "burn" amount if free+reserved do not suffice.

pallets/tokens/src/lib.rs
```diff
@@ -875,3 +911,256 @@ where
...
+		<NextCurrencyId<T>>::mutate(|id| *id += One::one());
+        let tokens = <Self as MultiTokenCurrency<T::AccountId>>::issue(token_id, amount);
+        // we are creating new token so amount can not be overflowed as its always true
+        // 0 + amount < T::Balance::max_value()
+        let _ = <Self as MultiTokenCurrency<T::AccountId>>::resolve_into_existing(token_id, address, tokens);
+        token_id
+    }
+
+    fn mint(currency_id: T::CurrencyId, address: &T::AccountId, amount: T::Balance) -> DispatchResult {
+        let curren_balance =  <Self as MultiTokenCurrency<T::AccountId>>::total_balance(currency_id, address);
```
iStrike7, line 1153: curren_balance typo

pallets/tokens/src/lib.rs
```diff
@@ -186,6 +193,8 @@ decl_storage! {
 		///
 		/// NOTE: This is only used in the case that this module is used to store balances.
 		pub Accounts get(fn accounts): double_map hasher(blake2_128_concat) T::AccountId, hasher(twox_64_concat) T::CurrencyId => AccountData<T::Balance>;
+
+        NextCurrencyId get(fn next_asset_id): T::CurrencyId;
```
iStrike7, line 197: maybe once this PR is verified, we can run cargo fmt to fix such indentation errors, before merging.

pallets/tokens/src/multi_token_currency.rs
```diff
@@ -0,0 +1,317 @@
...
+	fn remove_lock(
+        currency_id: Self::CurrencyId,
+		id: LockIdentifier,
+		who: &AccountId,
+	);
+}
+
+pub trait MultiTokenCreatableCurrency<AccountId>: MultiTokenCurrency<AccountId> {
+	fn create(address: &AccountId, amount: Self::Balance) -> Self::CurrencyId;
+    fn mint(currency_id: Self::CurrencyId, address: &AccountId, amount: Self::Balance) -> DispatchResult;
```
iStrike7, line 315: I was thinking we'd use separate traits for creating and minting.
So we'd have MultiTokenCreatableCurrency and MultiTokenMintableCurrency.
Not sure if that's required though, maybe we could rename MultiTokenCreatableCurrency as MultiTokenCurrencyExtended.

pallets/tokens/src/lib.rs
```diff
@@ -875,3 +911,256 @@ where
...
+impl<T> MultiTokenCreatableCurrency<T::AccountId> for MultiTokenCurrencyAdapter<T> where
+    T: Trait
+{
+	fn create(address: &T::AccountId, amount: T::Balance) -> T::CurrencyId{
+        let token_id = <NextCurrencyId<T>>::get();
+		<NextCurrencyId<T>>::mutate(|id| *id += One::one());
+        let tokens = <Self as MultiTokenCurrency<T::AccountId>>::issue(token_id, amount);
+        // we are creating new token so amount can not be overflowed as its always true
+        // 0 + amount < T::Balance::max_value()
+        let _ = <Self as MultiTokenCurrency<T::AccountId>>::resolve_into_existing(token_id, address, tokens);
```
iStrike7, line 1148: shouldn't this be resolve_creating instead?

pallets/tokens/src/lib.rs
```diff
@@ -875,3 +911,256 @@ where
...
+        token_id
+    }
+
+    fn mint(currency_id: T::CurrencyId, address: &T::AccountId, amount: T::Balance) -> DispatchResult {
+        let curren_balance =  <Self as MultiTokenCurrency<T::AccountId>>::total_balance(currency_id, address);
+        // check for overflow while minting
+        let _ = curren_balance.checked_add(&amount).ok_or(Error::<T>::BalanceOverflow)?;
+        let tokens = <Self as MultiTokenCurrency<T::AccountId>>::issue(currency_id, amount);
+        // we are always sure that there will be no overflow as it was checked
+        let _ = <Self as MultiTokenCurrency<T::AccountId>>::resolve_into_existing(currency_id, address, tokens);
```
iStrike7, line 1158: shouldn't this too be resolve_creating instead?

pallets/tokens/src/lib.rs
iStrike7: Even though I believe they are both are functionally the same as far as the currency ORML tokens are concerned.

runtime/src/lib.rs
```diff
@@ -496,7 +508,7 @@ construct_runtime!(
         Balances: pallet_balances::{Module, Call, Storage, Config<T>, Event<T>},
         TransactionPayment: pallet_transaction_payment::{Module, Storage},
         Sudo: pallet_sudo::{Module, Call, Config<T>, Storage, Event<T>},
-        Assets: pallet_assets::{Module, Call, Storage, Event<T>},
+        // Assets: pallet_assets::{Module, Call, Storage, Event<T>},
```
iStrike7, line 511: maybe you can just remove this.

pallets/xyk/src/lib.rs
```diff
@@ -139,47 +154,48 @@ decl_module! {
...
-
-            <assets::Module<T>>::assets_transfer(
-                &first_asset_id,
+            T::Currency::transfer(
+                first_asset_id,
                 &sender,
                 &vault,
-                &first_asset_amount
+                first_asset_amount,
+                ExistenceRequirement::KeepAlive
```
iStrike7, line 168: maybe change all ExistenceRequirement to AllowDeath, as transfer in ORML does allow death

pallets/xyk/src/lib.rs
```diff
@@ -117,12 +130,12 @@ decl_module! {
...
 
             ensure!(
-                <assets::Module<T>>::balance(first_asset_id, sender.clone()) >= first_asset_amount,
+                T::Currency::free_balance(first_asset_id, &sender) >= first_asset_amount,
                 Error::<T>::NotEnoughAssets,
             );
 
             ensure!(
-                <assets::Module<T>>::balance(second_asset_id, sender.clone()) >= second_asset_amount,
+                T::Currency::free_balance(second_asset_id, &sender) >= second_asset_amount,
```
iStrike7, line 138: why not use ensure_can_withdraw? as it also checks for frozen amount.

pallets/xyk/src/lib.rs
iStrike7: not that it matters as the argument in ignored in ORML.

pallets/xyk/src/lib.rs
iStrike7: and these checks have to made even though transfer itslef calls ensure_can_withdraw, as we are making two calls to transfer and if one passes the other shouldn't fail. So we need to ensure that both pass.

pallets/xyk/src/lib.rs
```diff
@@ -139,47 +154,48 @@ decl_module! {
...
-            min_amount_out: T::Balance,
+            sold_asset_id: CurrencyIdOf<T>,
+            bought_asset_id: CurrencyIdOf<T>,
+            sold_asset_amount: BalanceOf<T>,
+            min_amount_out: BalanceOf<T>,
         ) -> DispatchResult {
 
             let sender = ensure_signed(origin)?;
 
+            debug::native::error!("hwllo world");
```
iStrike7, line 198: remove this maybe?

pallets/xyk/src/lib.rs
```diff
@@ -420,7 +442,7 @@ decl_module! {
             let liquidity_asset_id = Self::get_liquidity_asset(first_asset_id, second_asset_id);
 
             ensure!(
-                <assets::Module<T>>::balance(liquidity_asset_id, sender.clone()) >= liquidity_asset_amount,
+                T::Currency::can_slash(liquidity_asset_id, &sender, liquidity_asset_amount),
```
iStrike7, line 445: I think we also need to check ensure_can_withdraw, as I believe that when burning frozen amount should be considered.

pallets/erc20-app/src/lib.rs
iStrike7: I think we also need to check ensure_can_withdraw, as I believe that when burning frozen amount should be considered.

pallets/eth-app/src/lib.rs
```diff
@@ -86,11 +89,21 @@ decl_module! {
             let who = ensure_signed(origin)?;
             let asset_id: BridgedAssetId = H160::zero();
 
-            // <asset::Module<T>>::do_burn(asset_id, &who, amount)?;
             let asset_id = <asset::Module<T>>::get_native_asset_id(asset_id);
-            let result = <assets::Module<T>>::assets_burn(&asset_id, &who, &amount.low_u128().saturated_into::<T::Balance>());
 
-            ensure!(result.is_ok(), Error::<T>::BurnFailure);
+            ensure!(T::Currency::can_slash(
```
iStrike7, line 94: I think we also need to check ensure_can_withdraw, as I believe that when burning frozen amount should be considered.

pallets/xyk/src/lib.rs
```diff
@@ -453,15 +476,15 @@ decl_module! {
                 second_asset_reserve - second_asset_amount,
             );
 
-            if (first_asset_reserve - first_asset_amount == 0.saturated_into::<T::Balance>())
-                || (second_asset_reserve - second_asset_amount == 0.saturated_into::<T::Balance>()) {
+            if (first_asset_reserve - first_asset_amount == 0.saturated_into::<BalanceOf<T>>())
+                || (second_asset_reserve - second_asset_amount == 0.saturated_into::<BalanceOf<T>>()) {
```
iStrike7, line 480: IMP!!
This may not be a part of modifications related to Tokens, but shouldn't LiquidityAssets and LiquidityPools also be removed?
Moreover, wouldn't zero balance in reserve cause divide by zero error in mint_liquidity function?
We could keep the pool alive by keeping LiquidityAssets and LiquidityPools, if we take measures to avoid divide by zero in the entire pallet. Which we should do any way as divide by zero might happen even if just one of the assets drops to zero.

iStrike7: Also shouldn't mint_liquidity also use liquidity instead of first_asset_amount for symmetry?

pallets/xyk/src/lib.rs
```diff
@@ -470,10 +493,10 @@ decl_module! {
 
 impl<T: Trait> Module<T> {
     pub fn calculate_sell_price(
-        input_reserve: T::Balance,
-        output_reserve: T::Balance,
-        sell_amount: T::Balance,
-    ) -> T::Balance {
+        input_reserve: BalanceOf<T>,
```
iStrike7, line 496: These functions also seem to be susceptible to divide by zero.
We need to check the entire pallet for divide by zero.

pallets/xyk/src/lib.rs
```diff
@@ -503,25 +526,26 @@ impl<T: Trait> Module<T> {
...
-        first_asset_id: T::AssetId,
-        second_asset_id: T::AssetId,
-    ) -> T::AssetId {
+        first_asset_id: CurrencyIdOf<T>,
+        second_asset_id: CurrencyIdOf<T>,
+    ) -> CurrencyIdOf<T> {
+        //TODO get rid of unwraps
         if <LiquidityAssets<T>>::contains_key((first_asset_id, second_asset_id)) {
-            <LiquidityAssets<T>>::get((first_asset_id, second_asset_id))
+            <LiquidityAssets<T>>::get((first_asset_id, second_asset_id)).unwrap()
```
iStrike7, line 538: I do not think we should use unwrap(), as get_liquidity_asset might get called without checking if LiquidityAssets contains key (either way)

iStrike7: Also in some places to check if the pool exists Pools is used and in others LiquidityAssets is used, I'm not sure if that's the right approach.

pallets/xyk/src/lib.rs
```diff
@@ -503,25 +526,26 @@ impl<T: Trait> Module<T> {
...
         } else {
-            <LiquidityAssets<T>>::get((second_asset_id, first_asset_id))
+            <LiquidityAssets<T>>::get((second_asset_id, first_asset_id)).unwrap()
```
iStrike7, line 540: I do not think we should use unwrap(), as get_liquidity_asset might get called without checking if LiquidityAssets contains key (either way)

pallets/xyk/src/mock.rs
```diff
@@ -56,18 +57,34 @@ impl system::Trait for Test {
...
-    type Balance = u128;
-    type AssetId = u32;
+impl Trait for Test {
     type Event = ();
+    type Currency = MultiTokenCurrencyAdapter<Test>;
 }
 
 pub type XykStorage = Module<Test>;
 
+impl<T: Trait> Module<T> {
```
iStrike7, line 76: why is this required?

runtime/src/lib.rs
```diff
@@ -475,6 +473,20 @@ impl pallet_assets_info::Trait for Runtime {
...
+    type Event = Event;
+    type Balance = Balance;
+    type Amount = i128;
+    type CurrencyId = u32;
+    type OnReceived = ();
+    type WeightInfo = ();
+}
+
+parameter_types! {
+    pub const GetNativeCurrencyId: u32 = 0;
```
iStrike7, line 489: where is this being used?

runtime/src/lib.rs
```diff
@@ -475,6 +473,20 @@ impl pallet_assets_info::Trait for Runtime {
     type MinLengthDescription = MinLengthDescription;
     type MaxLengthDescription = MaxLengthDescription;
     type MaxDecimals = MaxDecimals;
+    type Currency = orml_tokens::MultiTokenCurrencyAdapter<Runtime>;
+}
+
+impl orml_tokens::Trait for Runtime {
+    type Event = Event;
+    type Balance = Balance;
+    type Amount = i128;
```
iStrike7, line 482: Maybe define this in the same place Balance is defined?

iStrike7 (changes requested):
Some changes required.

pallets/tokens/src/lib.rs
mateuszaaa: i would not format external crates as it might be hard in the future to upgrade to newer version (because of the conflicts) - for our code i have just applied formatting

mateuszaaa: probably would bo good to have some github check that enforces that :thinking:

pallets/tokens/src/lib.rs
```diff
@@ -875,3 +911,256 @@ where
...
+
+	fn can_slash(currency_id: T::CurrencyId, who: &T::AccountId, value: Self::Balance) -> bool {
+		Module::<T>::can_slash(currency_id, who, value)
+	}
+
+	fn total_issuance(currency_id: T::CurrencyId) -> Self::Balance {
+		Module::<T>::total_issuance(currency_id)
+	}
+
+	fn minimum_balance() -> Self::Balance {
```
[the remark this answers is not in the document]
mateuszaaa: hmmmm, balance seems to be independent from currency - so you see any future use cases for that ?

pallets/tokens/src/lib.rs
mateuszaaa: eventualy i used deposit_creating that works with `Balance` instead of `Imbalance`

pallets/tokens/src/lib.rs
mateuszaaa: ok, i changed it a little bit. I think interface should not allow for minting tokens that does not exists - instead issue should be used. @iStrike7 please take a look on updated impl

pallets/xyk/src/lib.rs
mateuszaaa: is that expected behaviour to remove account when balance is 0 ?

pallets/xyk/src/mock.rs
mateuszaaa: just to keep test code a bit simpler/readable without need to cast type N times in single test case scenario

runtime/src/lib.rs
mateuszaaa: some leftovers...

pallets/xyk/src/lib.rs
```diff
@@ -346,36 +379,40 @@ decl_module! {
...
 
             ensure!(
-                <assets::Module<T>>::balance(second_asset_id, sender.clone()) >= second_asset_amount,
+                T::Currency::free_balance(second_asset_id, &sender) >= second_asset_amount,
                 Error::<T>::NotEnoughAssets,
             );
 
-            <assets::Module<T>>::assets_transfer(
-                &first_asset_id,
+            T::Currency::mint(liquidity_asset_id,&sender, liquidity_assets_minted)?;
```
gleb-urvanov, line 406: move minting after provided assets transfer

pallets/xyk/src/lib.rs
gleb-urvanov: The idea was this:
If Pools are removed, the `mint_liquidity` is no longer callable for the pool. Instead, the pool can be recreated using `create_pool`. In this case, LiquidityAssets and LiquidityPools for the corresponding pool will be reinitialized with new values. This also means divizion by zero will not happen.

gleb-urvanov: as for using liquidity for minting- unfortunately we had to drop the idea of having mint and burn symmetrical because of the different semantics:
in both cases, the user calling the function is aiming to give some assets they possess in exchange for another asset. Unlike trading, is most cases they will be aiming to provide the exact amount of assets. Forcing them to provide a destination amount may result in unexpected amount to be taken from the user if some thirdparty transaction will affect the balance in the pool

gleb-urvanov (changes requested):

pallets/assets-info/src/lib.rs
```diff
@@ -46,8 +47,14 @@ pub trait Trait: assets::Trait {
 
     /// The maximum decimal points an asset may be.
     type MaxDecimals: Get<u32>;
+
+    type Currency: MultiTokenCurrencyExtended<Self::AccountId>;
 }
 
+type CurrencyIdOf<T> = <<T as Trait>::Currency as MultiTokenCurrency<
```
gleb-urvanov, line 54: Are there a point in recreating `CurrencyIdOf` and `BalanceOf` types in every pallet where these are used?

pallets/tokens/src/lib.rs
```diff
@@ -875,3 +911,258 @@ where
...
+	}
+
+	fn mint(currency_id: T::CurrencyId, address: &T::AccountId, amount: T::Balance) -> DispatchResult {
+		if ! Self::exists(currency_id){
+			return Err(DispatchError::from(Error::<T>::TokenIdNotExists));
+		}
+		let current_balance =  <Self as MultiTokenCurrency<T::AccountId>>::total_balance(currency_id, address);
+		// check for overflow while minting
+		current_balance.checked_add(&amount).ok_or(Error::<T>::BalanceOverflow)?;
+		let tokens = <Self as MultiTokenCurrency<T::AccountId>>::issue(currency_id, amount);
```
mateuszaaa, line 1158: should be resolve/deposit into creating

mateuszaaa, line 1158: tbr

pallets/tokens/src/lib.rs
```diff
@@ -875,3 +911,258 @@ where
...
+		// we are always sure that there will be no overflow as it was checked
+		let _ = <Self as MultiTokenCurrency<T::AccountId>>::resolve_into_existing(currency_id, address, tokens);
+		Ok(())
+	}
+
+	fn exists(currency_id: Self::CurrencyId) -> bool{
+		<TotalIssuance<T>>::contains_key(currency_id)
+	}
+
```
mateuszaaa, line 1167: burning should be only performing free balance - maybe burn function should be introduced ??

pallets/eth-app/src/lib.rs
```diff
@@ -86,12 +91,32 @@ decl_module! {
...
+                Error::<T>::BurnFailure);
+
+            let new_balance = T::Currency::free_balance(asset_id, &who) - amount;
+            T::Currency::ensure_can_withdraw(asset_id,
+                &who,
+                amount,
+                WithdrawReasons::all(),
+                new_balance).or(Err(Error::<T>::BurnFailure))?;
+
+            T::Currency::slash(
```
mateuszaaa, line 113: introduct burn&settle function instead

mateuszaaa: should not use burn directly as its not working properly when one wants to burn more tokens that available

pallets/erc20-app/src/lib.rs
```diff
@@ -87,17 +93,35 @@ decl_module! {
...
+
+            ensure!(T::Currency::ensure_can_withdraw(
+                native_asset_id,
+                &who,
+                amount,
+                WithdrawReasons::all(),
+                new_balance).is_ok(),
+                Error::<T>::BurnFailure);
+
+            T::Currency::slash(
```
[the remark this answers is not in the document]
mateuszaaa: should be replaced by burn&settle

pallets/tokens/src/multi_token_currency.rs
```diff
@@ -0,0 +1,317 @@
...
+	/// Remove an existing lock.
+	fn remove_lock(
+        currency_id: Self::CurrencyId,
+		id: LockIdentifier,
+		who: &AccountId,
+	);
+}
+
+pub trait MultiTokenCurrencyExtended<AccountId>: MultiTokenCurrency<AccountId> {
+	fn create(address: &AccountId, amount: Self::Balance) -> Self::CurrencyId;
```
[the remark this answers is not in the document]
mateuszaaa: there is issue function already in origin Currency trait

pallets/eth-app/src/lib.rs
mateuszaaa: good reference: `MultiCurrency::withdraw`

pallets/assets-info/src/lib.rs
mateuszaaa: it doesnt seem to be workaround for that as we are always casting to `Trait` type that is availble inside given pallet

pallets/xyk/src/lib.rs
mateuszaaa: agreed to change to `AllowDeath`

pallets/xyk/src/lib.rs
mateuszaaa: https://trello.com/c/SJ43cop3/105-figure-out-if-remove-liquiditypools-along-with-pools

pallets/assets-info/src/lib.rs
mateuszaaa: https://trello.com/c/OST4jO2F/106-create-single-trait-that-sets-currencyid-and-balance-and-tightly-couple-it-with-all-the-pallets-to-have-common-type-in-typescrip

pallets/xyk/src/lib.rs
```diff
@@ -139,43 +164,43 @@ decl_module! {
                 (second_asset_id, first_asset_id), second_asset_amount
             );
 
-            let liquidity_asset_id = <assets::Module<T>>::assets_next_asset_id();
+
+            let liquidity_asset_id = T::Currency::create(&sender, initial_liquidity);
```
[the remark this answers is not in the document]
mateuszaaa: but how to transfer assets while `liquidity_asset_id` is not known ? Now how asset id are generated is hidden from the user - is reordering a significant problem here?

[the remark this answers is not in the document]
gleb-urvanov: Why do we need to know liquidity asset id to transfer non-liquidity assets?
The reordering is not a big issue, just makes more sense in this order- first we take assets from the user, and then provide them with liquidity asset.
What I'm suggesting here is just to move liquidity assets creation and storing liquidity pools and liquidity assets to the line 189.

pallets/tokens/src/lib.rs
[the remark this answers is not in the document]
iStrike7: Yeah, some currencies may require a non-zero minimum balance for some reason. You can just add it as an argument, but still just return Zero::zero for now

pallets/tokens/src/lib.rs
```diff
@@ -875,3 +911,270 @@ where
...
+		let _ = <Self as MultiTokenCurrency<T::AccountId>>::deposit_creating(currency_id, address, amount);
+		Ok(())
+	}
+
+	fn exists(currency_id: Self::CurrencyId) -> bool{
+		<TotalIssuance<T>>::contains_key(currency_id)
+	}
+
+	/// either succeeds or leaves state unchanged
+	fn burn_and_settle(currency_id: T::CurrencyId, who: &T::AccountId, amount: T::Balance) -> DispatchResult{
```
iStrike7, line 1168: This is the implemntation i had in mind:

```
fn withdraw(currency_id: Self::CurrencyId, who: &T::AccountId, amount: Self::Balance) -> DispatchResult {
		if amount.is_zero() {
			return Ok(());
		}
		Self::ensure_can_withdraw(currency_id, who, amount)?;
		<TotalIssuance<T>>::mutate(currency_id, |v| *v -= amount);
		Self::set_free_balance(currency_id, who, Self::free_balance(currency_id, who) - amount);

		Ok(())
	}
```

pallets/xyk/src/lib.rs
```diff
@@ -503,53 +551,62 @@ impl<T: Trait> Module<T> {
...
     pub fn get_liquidity_asset(
-        first_asset_id: T::AssetId,
-        second_asset_id: T::AssetId,
-    ) -> T::AssetId {
-        if <LiquidityAssets<T>>::contains_key((first_asset_id, second_asset_id)) {
-            <LiquidityAssets<T>>::get((first_asset_id, second_asset_id))
-        } else {
-            <LiquidityAssets<T>>::get((second_asset_id, first_asset_id))
-        }
+        first_asset_id: CurrencyIdOf<T>,
```
iStrike7, line 558: I do not think that we should use this approach, as get I think will return the default value if it doesn't find anything. So we definitely can't use that.

pallets/tokens/src/lib.rs
mateuszaaa: actually once i realized that we can achive the same (i think) behaviour using existing methods i thought its will be more "proper way". But this way its definately easier to understand :+1:

gleb-urvanov (approved):

iStrike7 (approved):

PR #41, opened 2021-05-05
Feature/extrinsic deserialization

mateuszaaa:
still TODO:
- ~fix bug - information about extrinsic (signer,nounce) is removed before extrincs are being executed (and shuffled)~
- ~improve randomization - currently only groups are randomized (single group per single extrincis signer)~
- ~remove extrinsic information when no longer needed~

### DIFF
--- a/client/block-builder/src/lib.rs
+++ b/client/block-builder/src/lib.rs
@@ -39,17 +39,14 @@ use sp_api::{
 };
 use sp_blockchain::{Backend, Error};
 use sp_consensus::RecordProof;
-use sp_core::ExecutionContext;
+use sp_core::{ExecutionContext, H256};
 use sp_runtime::{
 	generic::BlockId,
 	traits::{
 		BlakeTwo256, Block as BlockT, DigestFor, Hash, HashFor, Header as HeaderT, NumberFor, One,
 	},
 };
-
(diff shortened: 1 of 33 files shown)


gleb-urvanov:
Do we have any alterations in `client/transaction-pool`? What are these?

client/block-builder/src/lib.rs
```diff
@@ -206,21 +213,22 @@ where
 	/// The storage proof will be `Some(_)` when proof recording was enabled.
 	pub fn build(
 		mut self,
-	) -> Result<BuiltBlock<Block, backend::StateBackendFor<B, Block>>, ApiErrorFor<A, Block>> {
+		context_provider: &dyn Fn(H256) -> Option<(AccountId32, u32)>
```
gleb-urvanov, line 216: totally missing where the `context_provider` is coming from, can you explain where it is implemented?

mateuszaaa:
@gleb-urvanov i tried to separate forking & modifications into separate commits

https://github.com/mangata-finance/mangata-node/compare/4101cd9f38dfb6e4a714850988e1126b1b82c936...e734429bfe0b4c4c3daaaf2ab3ae90d538350b75

PR #42, opened 2021-05-06
Feature/tokens mint and create only as root

mateuszaaa:


pallets/tokens/src/lib.rs
```diff
@@ -325,12 +325,12 @@ decl_module! {
             token_id: TokenId,
             account_id: T::AccountId,
             value: Balance,
-        ) -> frame_support::dispatch::DispatchResult{
```
gleb-urvanov, line 328: does it makes sense to remove `DispatchResult` ?

mateuszaaa: i think thats the default one if not specified manually

gleb-urvanov (approved):

PR #43, opened 2021-05-07
Feature/calculate with id

stanieltron:
added pub fns and rpc call to calculate buy and sell price providing token_id instead of reserves


pallets/xyk/src/lib.rs
```diff
@@ -546,6 +546,48 @@ impl<T: Trait> Module<T> {
...
+            .saturated_into::<u128>()
+            .into();
+        let output_reserve_u256: U256 = Pools::get((bought_token_id, sold_token_id))
+            .saturated_into::<u128>()
+            .into();
+        let sell_amount_u256: U256 = sell_amount.saturated_into::<u128>().into();
+
+        let input_amount_with_fee: U256 = sell_amount_u256 * 997;
+        let numerator: U256 = input_amount_with_fee * output_reserve_u256;
+        let denominator: U256 = input_reserve_u256 * 1000 + input_amount_with_fee;
```
gleb-urvanov, line 564: Could calculations be abstracted from the method? Currently we have price calculation code duplication, which is a significant flaw.
We should have only two source-of-truth function calculating price- one for sell price, one for buy price. The rest of the function should not know how the price is calculated.

pallets/xyk/src/lib.rs
```diff
@@ -546,6 +546,48 @@ impl<T: Trait> Module<T> {
...
+            .saturated_into::<u128>()
+            .into();
+        let output_reserve_u256: U256 = Pools::get((bought_token_id, sold_token_id))
+            .saturated_into::<u128>()
+            .into();
+        let buy_amount_u256: U256 = buy_amount.saturated_into::<u128>().into();
+
+        let numerator: U256 = input_reserve_u256 * buy_amount_u256 * 1000;
+        let denominator: U256 = (output_reserve_u256 - buy_amount_u256) * 997;
+        let result: U256 = numerator / denominator + 1;
```
gleb-urvanov, line 586: Same as above comment

gleb-urvanov (changes requested):

gleb-urvanov (changes requested):
Revert style alterations

pallets/xyk/src/lib.rs
```diff
@@ -463,16 +465,19 @@ impl<T: Trait> Module<T> {
...
         let output_reserve_saturated: U256 = output_reserve.saturated_into::<u128>().into();
         let buy_amount_saturated: U256 = buy_amount.saturated_into::<u128>().into();
 
         let numerator: U256 = input_reserve_saturated * buy_amount_saturated * 1000;
         let denominator: U256 = (output_reserve_saturated - buy_amount_saturated) * 997;
-        let result: U256 = numerator / denominator + 1;
+        let result = numerator
+            .checked_div(denominator)
+            .ok_or_else(|| DispatchError::from(Error::<T>::DivisionByZero))?
+            - 1;
```
gleb-urvanov, line 478: Here should be `+ 1` to compensate the rounding, not `- 1`

gleb-urvanov (changes requested):

gleb-urvanov (approved):

PR #51, opened 2021-05-21
fix for multivalidator mandatory set_uncles transaction failure

gleb-urvanov:


iStrike7:
Is this merge only to fork the authorship pallet?
And also shouldn't the pallet-authorship now be configured as a local dependency, in modules that require it as a dependency, such as the staking pallet and the runtime.

gleb-urvanov:
> Is this merge only to fork the authorship pallet?
> And also shouldn't the pallet-authorship now be configured as a local dependency, in modules that require it as a dependency, such as the staking pallet and the runtime.

There is a one-line change in `authorship/src/lib.rs:line213` under `// FIXME` comment. 
The `set_uncles` in no longer mandatory, and `DispatchClass` dependency removed as no longer used.

pallet-authorship is configured as a local dependency using patch mechanism in Cargo.toml. It rewrites the pallet-authorship dependency for every crate which requires it.

iStrike7 (approved):

mateuszaaa (approved):
LGTM seems like there is still some merge conflict

PR #54, opened 2021-05-31
Feature/seed in block

mateuszaaa:
- shuffle extrinsics using Fisher-Yates algorithm, Xoshiro256+ is used as pseudo random number source. It was required to replace origin shuffling implementation to something that can be reproducible in JS api
- seed used for Xoshiro256+ initialization is included in block as mandatory inherent/extrinsic

that PR can be tested with updated JS api introduced in https://github.com/mangata-finance/tx-randomization-demo/pull/3


client/basic-authorship/src/basic_authorship.rs
```diff
@@ -302,7 +282,30 @@ where
 
         self.transaction_pool.remove_invalid(&unqueue_invalid);
 
-        let (block, storage_changes, proof) = block_builder.build()?.into_inner();
+        let (seed,inherents) = block_builder.create_inherents(inherent_data)?;
+        for inherent in  inherents {
```
gleb-urvanov, line 286: Why moved here, after `remove_invalid` method called?

client/block-builder/src/lib.rs
```diff
@@ -169,20 +173,22 @@ where
     ///
     /// This will ensure the extrinsic can be validly executed (by executing it).
     pub fn push(&mut self, xt: <Block as BlockT>::Extrinsic) -> Result<(), ApiErrorFor<A, Block>> {
-        info!("Pushing transactions without execution");
-        self.extrinsics.push(xt);
+        let exts = &mut self.extrinsics;
+        exts.push(xt.clone());
```
gleb-urvanov, line 177: Why we no longer can use `self.extrinsics` here?

pallets/random-seed/README.md
```diff
@@ -0,0 +1,74 @@
+# Timestamp Module
```
gleb-urvanov, line 1: Readme does not match the content

pallets/random-seed/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+        )]
+        fn set(origin, seed: H256) {
+            <Self as Store>::Seed::put(seed);
+        }
+
+    }
+}
+
+decl_storage! {
+    trait Store for Module<T: Trait> as Timestamp {
```
gleb-urvanov, line 37: trait still reads `Timestamp`

pallets/random-seed/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl IsFatalError for RandomSeedInherentError {
+	fn is_fatal_error(&self) -> bool {
+        true
+	}
+}
+
+fn extract_inherent_data(data: &InherentData) -> Result<RandomSeedInherentType, RuntimeString> {
+    data.get_data::<RandomSeedInherentType>(&RANDOM_SEED_INHERENT_IDENTIFIER)
+        .map_err(|_| RuntimeString::from("Invalid timestamp inherent data encoding."))?
```
gleb-urvanov, line 80: error string reads "timestamp"

pallets/random-seed/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+// originally in sp-module
+pub type RandomSeedInherentType = H256;
+const RANDOM_SEED_INHERENT_IDENTIFIER: InherentIdentifier = *b"blckseed";
+
+#[derive(Encode, sp_runtime::RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Decode))]
+pub enum RandomSeedInherentError {
```
gleb-urvanov, line 55: why we need this error enum and following error implementations?

gleb-urvanov (changes requested):

client/basic-authorship/src/basic_authorship.rs
mateuszaaa: seed is calculated from the current block extrinsic. In the original implementation inherents are put into the block first and then all the transactions. But at the time when random seed inherentis is executed it needs to know seed already thats why i needed to reorder `<inherents><extrinsics>` into `<extrinsics><inherents>`

client/block-builder/src/lib.rs
mateuszaaa: leftover, fixed now

pallets/random-seed/src/lib.rs
mateuszaaa: this is how inherents works in substrate. You implement `ProvideInherent` trait for given pallet and expose Inherent features in runtime. Block builder operates on inherent data passed from babe, those data can be easily registered anywhere as those are just mappings 
`identifier -> data`  then later on in runtime pallets that implements ProvideInherent tries to fetch those 
`identifier -> data` mappings, deserialize stored information(`H256` seed in our case) and automatically triggers proper extrinsic ( storage `Seed::set` ) to store seed in blockchain. But such deserialization may fail thats why proper errors needs to be defined and triggered

client/basic-authorship/src/basic_authorship.rs
gleb-urvanov: Got it. I assume this could be altered again after implementing the signing/verifying approach for the seed.

gleb-urvanov (approved):

PR #57, opened 2021-06-01
implemented with multiply_by_rational and unspecified saturated_into

iStrike7:


pallets/xyk/src/lib.rs
```diff
@@ -1366,18 +1327,13 @@ impl<T: Trait> Valuate for Module<T> {
...
         let liquidity_token_reserve_u256: U256 =
-            liquidity_token_reserve.saturated_into::<u128>().into();
-        let mng_token_amount_u256 =
-            mng_token_reserve_u256 * liquidity_token_amount_u256 / liquidity_token_reserve_u256;
-        let mng_token_amount = mng_token_amount_u256
-            .saturated_into::<u128>()
-            .saturated_into::<Self::Balance>();
-        mng_token_amount
+            liquidity_token_reserve.into();
+        (mng_token_reserve_u256 * liquidity_token_amount_u256 / liquidity_token_reserve_u256).saturated_into()
```
gleb-urvanov, line 1336: why not `multiply_by_rational` here?

pallets/xyk/src/lib.rs
```diff
@@ -1388,15 +1344,11 @@ impl<T: Trait> Valuate for Module<T> {
...
-        let scaled_liquidity_token_amount_u256 =
-            liquidity_token_amount_u256 * mng_token_amount_u256 / mng_valuation_u256;
-        let scaled_liquidity_token_amount = scaled_liquidity_token_amount_u256
-            .saturated_into::<u128>()
-            .saturated_into::<Self::Balance>();
-        scaled_liquidity_token_amount
+            liquidity_token_amount.into();
+        let mng_token_amount_u256: U256 = mng_token_amount.into();
+
+        (liquidity_token_amount_u256 * mng_token_amount_u256 / mng_valuation_u256).saturated_into()
```
gleb-urvanov, line 1352: why not `multiply_by_rational` here?

pallets/xyk/src/lib.rs
iStrike7: multiply_by_rational returns an error if the result won't fit into u128.
Here although it seems clear that liquidity_token_amount_u256 / liquidity_token_reserve_u256 will always be less than one, since this function faces outside the pallet it may happen that it is called with inconsistent or de-synced data (Although I currently see no scenario where that is possible with staking, which is the pallet currently using this trait)
So to avoid raising a panic or deal with undefined behaviour ( given that I do not want to change the function signature to use errors, as the part of staking that uses this function does not handle errors gracefully) I do not use multiply_by_rational here.

gleb-urvanov (approved):

pallets/xyk/src/lib.rs
iStrike7: Same.

PR #58, opened 2021-06-01
Feature/mint slippage

stanieltron:
API CHANGE NEEDED:
Introduced mint slippage by adding expected_second_asset_amount fn argument, which is checked against real second asset amount. 

On FE side, FE needs to get current second_asset_amount required to mint and add 1% buffer. Use this value as  expected_second_asset_amount, as this will be compared to second_asset_amount required in block.

BEFORE:
fn mint_liquidity(
        sender: AccountId,
        first_asset_id: Self::CurrencyId,
        second_asset_id: Self::CurrencyId,
        first_asset_amount: Self::Balance,
    ) -> DispatchResult;

NOW:
fn mint_liquidity(
        sender: AccountId,
        first_asset_id: Self::CurrencyId,
        second_asset_id: Self::CurrencyId,
        first_asset_amount: Self::Balance,
        expected_second_asset_amount: Self::Balance,
    ) -> DispatchResult;

New ensure and error added:
ensure!(
            second_asset_amount <= expected_second_asset_amount,
            Error::<T>::SecondAssetAmountExceededExpectations,
        );
        
Unit test added


pallets/xyk/src/lib.rs
```diff
@@ -395,11 +396,12 @@ decl_module! {
             first_asset_id: TokenId,
             second_asset_id: TokenId,
             first_asset_amount: Balance,
+            expected_second_asset_amount: Balance,
```
gleb-urvanov, line 399: Here and everywhere new parameter was added:
Make new parameter optional to avoid breaking the api

pallets/xyk/src/lib.rs
```diff
@@ -1025,6 +1029,11 @@ impl<T: Trait> XykFunctionsTrait<T::AccountId> for Module<T> {
             .saturated_into::<u128>()
             .saturated_into::<Self::Balance>();
 
+        ensure!(
+            second_asset_amount <= expected_second_asset_amount * 101 / 100,
```
gleb-urvanov, line 1033: why 1% is used here instead of a plain comparison?

gleb-urvanov (changes requested):
Also document api changes in the PR description

gleb-urvanov (approved):

PR #59, opened 2021-06-01
Feature/trigger e2e tests

goncer:
Changes done:
* Workflow can skip some steps based on the PR labels [skip-build, skip-e2e-tests, skip-publish]
** Right now, I am enforcing running the tests to publish, but can be changed.
* The Action now has 3 differentiate phases and based on the skips, will avoid some of the steps.
* Now its pointing to a e2e branch , so I would merge first the e2e and then fix the ref. on this workflow.
Wiki!
https://github.com/mangata-finance/wiki/wiki/CI---CD-Workflow


.github/workflows/docker-image.yml
```diff
@@ -1,17 +1,130 @@
...
+        if: "!contains(github.event.pull_request.labels.*.name, 'skip-build')"
+        run: |
+          docker load --input /tmp/${{ github.run_id }}.tar
+          docker image ls -a
+      - 
+        name: Checkout tests
+        uses: actions/checkout@v2
+        with:
+          repository: mangata-finance/mangata-e2e
+          token: ${{ secrets.E2E_GITHUB_CHECKOUT_TOKEN }}
```
gleb-urvanov, line 90: the token is no longer necessary

gleb-urvanov (changes requested):

.github/workflows/docker-image.yml
goncer: fixed.

gleb-urvanov (approved):

PR #68, opened 2021-06-13
done.

iStrike7:


gleb-urvanov (approved):
Please document the asset info format for the liquidity token (wiki is the right place to do it)

PR #70, opened 2021-06-14
Documenation of code alteration

mateuszaaa:


Mangata.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+| Alice TX1 |  Bob TX1  |  Alice TX2 |  Bob TX2 | Alice TX3 |
+```
+
+As shuffling occurs on block execution not creation, every node needs to calculate the same order and agree on storage state afterwards. That is achieved using Fisher-Yates shuffling algorithm with Xoshiro256++ as PRNG initialized with seed stored in blockchain storage.  Usage of sr25519 key pair together with VRF capabilities guarantees that block producer cannot impact execution order and it's easy to validate if the seed was properly calculated by validators/following nodes (thanks to VRF verification that only requires block author public key). Seed itself is stored into the storage and is publically accessible through dedicated runtime API(TODO PUT NAME HERE).
+
+PRNG initialization seed for block `N-1` is calculated as:
+- Seed from block `N-1`
+- babe epoch randomness (changed in every epoch)
```
gleb-urvanov, line 25: Isn't the randomness used in the first block of the epoch only? 
Isn't the seed encrypted?

gleb-urvanov (changes requested):

Mangata.md
mateuszaaa: >Isn't the randomness used in the first block of the epoch only?
no - i wasnt aware that we want to use it in that  fashion 
> Isn't the seed encrypted?
fixed

mateuszaaa:
@iStrike7 please take a look on staking pallet modification description

Mangata.md
```diff
@@ -0,0 +1,66 @@
...
+- [sc-basic-authorship](https://docs.rs/sc-basic-authorship/0.8.0/sc_basic_authorship/index.html) - shuffling extrinsics before passing them to block builder
+- [sc-consensus-babe](https://docs.rs/sc-basic-authorship/0.8.0/sc_basic_authorship/index.html) - calcuating and injecting shuffling seed value into InherentData, shuffling seed verification
+- [sc-service](https://docs.rs/sc-service/0.8.0/sc_service/index.html) - fetching seed value and extrinsic shuffling for 'following nodes'
+#Tokens
+
+Mangata uses concept of liquidity tokens. For every pair of token provided to the exchange(for staking purposes) a new liquidity token with unique ID is created. New currency trait [`MultiTokenCurrency`](https://github.com/mangata-finance/mangata-node/blob/0846c42a7b7fd29e19fd1b30043ddb3b55a8f250/pallets/tokens/src/multi_token_currency.rs#L14) was introduced  and integrated with staking(TODO link) pallet. Origin implementation was not sufficient as it only works with single, statically defined currency in our case we require supporting multiple dynamically created currencies.
```
iStrike7, line 39: "For every pair of token provided to the exchange(for staking purposes) a new liquidity token with unique ID is created."
It works more like. "You can create a liquidity pool using a pair of tokens, and then use the tokens of that pool to stake"

Mangata.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+Affected crates
+- [orml-tokens](https://docs.rs/orml-tokens/0.3.1/orml_tokens/index.html) - Introduced new `MultiTokenCurrency` trait, using it as `Ballances` replacement
```
iStrike7, line 43: typo

iStrike7: Ballances

gleb-urvanov (approved):

PR #74, opened 2021-06-21
create intermediate image before building the node and execute pre ch…

mateuszaaa:


.github/workflows/docker-image.yml
```diff
@@ -17,8 +17,45 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Build
+        uses: docker/build-push-action@v2
+        with:
+          file: "devops/dockerfiles/node-and-runtime/Dockerfile"
+          load: true
+          tags: mangatasolutions/mangata-node-ut:${{ github.sha }}
+          target: env_builder
```
goncer, line 52: just a question ( not sure either)
the sentence `target: env_builder ` is going to avoid building the image in both steps ? ie, it will first build it in :   `rustfmt-check` and later reuse it in `unit-test `?

I workaround building the image twice using, 
`outputs: type=docker,dest=/tmp/${{ github.run_id }}.tar`
Your solution seems to be cleaner!

goncer: + did you check that the above is compatible with the load image in the e2e tests? L112 We may need to adapt those also.
( Ideally we do not build the image more than one time )

 ```
       name: Load Docker image
        if: "!contains(github.event.pull_request.labels.*.name, 'skip-build')"
        run: |
          docker load --input /tmp/${{ github.run_id }}.tar
          docker image ls -a
```

mateuszaaa: yeah i saw the approach with loading images but building `env_builder` takes only 2 minutes so i decided to build it 3 times : 
- in `rust-fmt-check`
- in `unit-test` 
- in `build` maintaining the old approach with exporting/importing images 

Above targets are run in parallel - these way you dont have to check for `build` step to finish to check if you forgot to apply code formatting ;)

mateuszaaa: above are failing because code is not formatted and unit tests fails to compile - i have fixes for those in separate PR. Would be the best if we can approve this PR but i`ll merge it after proper fixes will be delivered to develop

goncer (approved):
looks good to me :)

PR #83, opened 2021-07-15
Shuffle extrinsics in runtime

mateuszaaa:
Temporary use one of the unused header fields in order to pass shuffling seed. Extending
sp_runtime::traits::Hash can be considered as one of the future options
for passing the seed


client/shuffler/src/lib.rs
```diff
@@ -1,17 +1,29 @@
...
+
+use sp_runtime::traits::{BlakeTwo256, Block as BlockT};
+use sp_runtime::AccountId32;
+use sp_std::collections::btree_map::BTreeMap;
+use sp_std::convert::TryInto;
+
+use sp_std::collections::vec_deque::VecDeque;
+use sp_std::vec::Vec;
+
+#[cfg(feature = "std")]
```
gleb-urvanov, line 13: Why do we need std here and later and will it work in WASM?

pallets/executive/src/lib.rs
```diff
@@ -329,7 +333,17 @@ where
             // execute extrinsics
             let (header, extrinsics) = block.deconstruct();
 
-            Self::execute_extrinsics_with_book_keeping(extrinsics, *header.number());
+            let extrinsics_with_author: Vec<(Option<_>,_)> = info.into_iter().zip(extrinsics.into_iter()).collect();
```
gleb-urvanov, line 336: Am I understand right the tx authorship info can not be accessed here inside executive using only data from `block`?

mateuszaaa: yep, only in this scope we have full information about all the types, inside ExecutivePallet these are not available

client/shuffler/src/lib.rs
mateuszaaa: this pallet needs to be compatible with/withour wasm because we are using it from the native part (block_budiled) and from inside the runtime(Executive pallet). So above is used to make it compatible with both. Its common approach from what i saw in other crates

gleb-urvanov (approved):

iStrike7 (approved):
Looks good.

PR #86, opened 2021-07-20
Feature/pay fees in mga

mateuszaaa:
Adaptations required for Treasury pallet to make it work with MGA token


iStrike7 (approved):
Looks good

PR #88, opened 2021-07-21
Fix for u128 input/output rpc paramters in XYK pallet API

mateuszaaa:
Actullay its a workaround for serde<->substrate issues.
Those seems to be fixed in v3.0.0. Trick is to use
NumberOrHex instead of u128 for input RPC api paramters.
NumberOrHex can be converted to u128 for XYK runtime api call
and is already shipped with proper serialization/deserialization
impl. What is more it allows for accepting request in both numerical
and hex/string form"

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8947,8 +8947,10 @@ dependencies = [
  "serde",
  "sp-api",
  "sp-blockchain",
+ "sp-core",
  "sp-rpc",
  "sp-runtime",
+ "sp-std",
  "xyk-runtime-api",
 ]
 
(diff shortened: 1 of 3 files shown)


iStrike7 (commented):
Awesome!
Is this tested with our current use cases and tests, both which were passing and the the ones which were failing?

mateuszaaa:
i did some testing, mostly with a little bit changed test case from ticket description as well as using `curl`. Can you point out some particular test case you want me to test ?

mateuszaaa:
together with @goncer we confirmed that fix is working

iStrike7 (approved):

PR #90, opened 2021-07-26
changed initial liquidity

stanieltron:
changed initial liquidity when creating the pool to prevent potential overflow. 

Previously initial liquidity amount = first asset amount + second asset amount.

Potential overflow in first and second asset amount are close to u128 max, overflowing at initial liquidity = ~ u128 max * 2

### DIFF
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -955,9 +955,7 @@ impl<T: Trait> XykFunctionsTrait<T::AccountId> for Module<T> {
         ensure!(first_asset_id != second_asset_id, Error::<T>::SameAsset,);
 
         // Liquidity token amount calculation
-        let initial_liquidity = first_asset_amount
-            .checked_add(second_asset_amount)
-            .ok_or_else(|| DispatchError::from(Error::<T>::MathOverflow))?;
+        let initial_liquidity = first_asset_amount;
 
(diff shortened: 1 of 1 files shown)


pallets/xyk/src/lib.rs
```diff
@@ -955,9 +955,7 @@ impl<T: Trait> XykFunctionsTrait<T::AccountId> for Module<T> {
         ensure!(first_asset_id != second_asset_id, Error::<T>::SameAsset,);
 
         // Liquidity token amount calculation
-        let initial_liquidity = first_asset_amount
-            .checked_add(second_asset_amount)
-            .ok_or_else(|| DispatchError::from(Error::<T>::MathOverflow))?;
+        let initial_liquidity = first_asset_amount;
```
goncer, line 958: @stanieltron , 
I think we have bunch of calculations based on this value. Are you sure that this will work? I'm not 100% confident, but some calculations assume that this liquidity amounts are nX+nY. Am I wrong?

ie:
burn _liquidity, L656,

`       
        let total_liquidity_assets: Balance =
            <T as Trait>::Currency::total_issuance(liquidity_asset_id.into()).into();
 let first_asset_amount = multiply_by_rational(
            first_asset_reserve,
            liquidity_asset_amount,
            total_liquidity_assets,
        )`

PR #93, opened 2021-07-28
Feature/initial liquidity asset amount change

stanieltron:
initial liquidity amount changed from

let initial_liquidity = first_asset_amount + second_asset_amount;

to

let initial_liquidity = first_asset_amount / 2 + second_asset_amount / 2;

### DIFF
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -955,9 +955,7 @@ impl<T: Trait> XykFunctionsTrait<T::AccountId> for Module<T> {
         ensure!(first_asset_id != second_asset_id, Error::<T>::SameAsset,);
 
         // Liquidity token amount calculation
-        let initial_liquidity = first_asset_amount
-            .checked_add(second_asset_amount)
-            .ok_or_else(|| DispatchError::from(Error::<T>::MathOverflow))?;
+        let initial_liquidity = first_asset_amount;
 
         Pools::insert((first_asset_id, second_asset_id), first_asset_amount);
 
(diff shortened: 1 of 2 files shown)


pallets/xyk/src/lib.rs
```diff
@@ -955,9 +955,7 @@ impl<T: Trait> XykFunctionsTrait<T::AccountId> for Module<T> {
         ensure!(first_asset_id != second_asset_id, Error::<T>::SameAsset,);
 
         // Liquidity token amount calculation
-        let initial_liquidity = first_asset_amount
-            .checked_add(second_asset_amount)
-            .ok_or_else(|| DispatchError::from(Error::<T>::MathOverflow))?;
+        let initial_liquidity = first_asset_amount / 2 + second_asset_amount / 2;
```
goncer, line 958: hey @stanieltron ,
how this is rounded?
Imagine the situation of a new pool ( 33333, 101010 ) -> initial_liquidity is:  67171.5.  That .5 is it rounded? 
My question goes in the direction of the following test scenario:
What happens if that decimal? the values will be accurate enough, so the right amount goes to the user and afterwards the pool can be burned completely?

gleb-urvanov: @goncer it is not rounded, integer arithmetic is in use here, so `/` is integer division operation. 33333 / 2 = 16666

goncer: Thanks for the anser @gleb-urvanov.
Already talked with @stanieltron , but for your information I found a small issue  ( actually automation did ) 
https://trello.com/c/EOzE0GCz
Basically creating up the smalled pool as possible ( 1X,1Y), makes people lose all the tokens they later invest in the pool, because 1/2 is -> 0, so the initial Liq token is 0.

gleb-urvanov: Great catch!

PR #95, opened 2021-08-02
Done and tested.

iStrike7:
Maciatko, Relay, Bob and Charlie and now initialized with MGA, along with Alice.


goncer:
Thanks Shoeb! :muscle:

PR #105, opened 2021-08-10
Feature/treasury fix

mateuszaaa:


iStrike7 (approved):
Not sure why this is required, but it'll do :)

PR #108, opened 2021-08-13
select e2e branch from param

goncer:
* with this change we can point to a specific e2e branch.


gleb-urvanov:
@goncer how to pass the parameter?

goncer:
> @goncer how to pass the parameter?

image

gleb-urvanov (approved):

PR #119, opened 2021-08-26
Removed docker hub runid tag

omaksi:


gleb-urvanov (commented):
Any reason why it causes the tests to fail?

goncer:
Ruinas is used to archive the generated image. I can fix it by replacing it
by sha.

goncer (changes requested):

goncer (approved):

PR #125, opened 2021-09-01
check if current branch is develop

goncer:


gleb-urvanov (approved):
LGTM however would wait for review from @omaksi before merging

omaksi:
But one thing that we could do is to refactor this script and break it into smaller ones

omaksi (approved):

goncer:
OK! 
Will merge it tomorrow and create a task to refactor it. 
Makes sense to split it. I think that the time has arrived :)

PR #126, opened 2021-09-03
simply check branch name

goncer:


goncer:
I think I have a simpler solution,
https://github.community/t/github-actions-branch-conditional/16057

gleb-urvanov (approved):

PR #148, opened 2021-11-25
Feature/encrypted transactions ut

iStrike7:


pallets/encrypted-transactions/src/mock.rs
```diff
@@ -0,0 +1,194 @@
...
+use sp_staking::SessionIndex;
+
+pub struct TestSessionManager;
+impl SessionManager<u64> for TestSessionManager {
+    fn end_session(_: SessionIndex) {}
+    fn start_session(_: SessionIndex) {}
+    fn new_session(_: SessionIndex) -> Option<Vec<u64>> {
+        None
+    }
+}
```
iStrike7, line 119: Where is this used?

pallets/encrypted-transactions/src/lib.rs
```diff
@@ -50,7 +55,7 @@ pub mod ecdsa {
 
 pub trait Trait: frame_system::Trait + pallet_session::Trait {
     type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
-    type Tokens: MultiTokenCurrency<Self::AccountId>;
+    type Tokens: MultiTokenCurrencyExtended<Self::AccountId>;
```
iStrike7, line 58: Not sure if we need to do this to use create in mock. But we can leave it like it is for now.

pallets/encrypted-transactions/src/mock.rs
```diff
@@ -0,0 +1,194 @@
...
+
+impl pallet_session::Trait for Test {
+    type Event = TestEvent;
+    type ValidatorId = <Self as frame_system::Trait>::AccountId;
+    type ValidatorIdOf = sp_runtime::traits::ConvertInto;
+    type ShouldEndSession = pallet_session::PeriodicSessions<(), ()>;
+    type NextSessionRotation = ();
+    type SessionManager = ();
+    type SessionHandler = TestSessionHandler;
```
iStrike7, line 159: This would push new keys for type `dummy: UintAuthorityId`. But later on we are using `crate::ecdsa::AuthorityId` for the pallet config.

1. How is this working?
2. Why not just work with `crate::ecdsa::AuthorityId` or `dummy: UintAuthorityId` all the way?

pallets/encrypted-transactions/src/lib.rs
```diff
@@ -123,7 +128,10 @@ decl_error! {
         CallDeserilizationFailed,
         DoublyEncryptedCallMaxLengthExceeded,
         TxnDoesNotExistsInRegistry,
-        UnexpectedError
+        UnexpectedError,
+        BalanceTooLowForFee,
+        TransactionAlreadyInQueue,
```
iStrike7, line 133: This does not seem to be used in lib.rs at all. Perhaps it should not be here.

pallets/encrypted-transactions/src/tests.rs
```diff
@@ -0,0 +1,390 @@
...
+        assert_err!(
+            EncryptedTX::submit_doubly_encrypted_transaction(
+                Origin::signed(1),
+                doubly_encrypted_call,
+                1,
+                1,
+                2,
+                3,
+            ),
+            Error::<Test>::TransactionAlreadyInQueue,
```
iStrike7, line 136: Shouldn't this be `TxnAlreadySubmitted`. How is this testing correctly?

pallets/encrypted-transactions/src/mock.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+use sp_runtime::testing::UintAuthorityId;
+use sp_runtime::traits::OpaqueKeys;
+
+pub struct TestSessionHandler;
+use sp_runtime::RuntimeAppPublic;
+
+impl SessionHandler<u128> for TestSessionHandler {
+    const KEY_TYPE_IDS: &'static [sp_runtime::KeyTypeId] = &[UintAuthorityId::ID];
```
iStrike7, line 128: Why is this defined here?

iStrike7 (commented):
This does not seem to test anything across sessions. Perhaps that would be a good addition.

PR #157, opened 2021-12-23
Assing Pays::No as default attribute for buy/sell asset extrinsics

mateuszaaa:
Fee will be still charged for transactions that fails


iStrike7 (approved):
Looks like Pays still has a similar syntax as in v2 for txn execution.
Thanks for looking into that!

PR #177, opened 2022-01-20
Feature/runtime seed validation

mateuszaaa:
- perform shuffling seed validation at runtime
- this check is executed by validator nodes


iStrike7 (approved):
LGTM

PR #178, opened 2022-01-20
Feature/liquidity mining rewards

stanieltron:
added functions for  minting liquidity rewards

NEW RPC:
 pub fn calculate_rewards_amount(
		user: AccountIdOf<T>,
		liquidity_asset_id: TokenId,
		block_number: u32,
	) -> Result<(Balance, i128) //(total earned, already claimed)
	
NEW EXTRINSIC:
fn claim_rewards(
		sender: AccountId,
		liquidity_token_id: Self::CurrencyId,
		amount: Self::Balance,
	)


pallets/xyk/src/lib.rs
```diff
@@ -347,6 +351,25 @@ pub mod pallet {
...
+
+	#[pallet::storage]
+	#[pallet::getter(fn liquidity_mining_user_claimed)]
+	pub type LiquidityMiningUserClaimed<T: Config> =
+		StorageMap<_, Blake2_256, (AccountIdOf<T>, TokenId), Option<i128>, ValueQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn liquidity_mining_pool_claimed)]
+	pub type LiquidityMiningPoolClaimed<T: Config> =
+		StorageMap<_, Blake2_256, TokenId, Option<Balance>, ValueQuery>;
```
mateuszaaa, line 372: not sure why keys are Option? Is there any use case for that ?

pallets/xyk/src/lib.rs
```diff
@@ -512,10 +535,366 @@ pub mod pallet {
...
 
 impl<T: Config> Pallet<T> {
+	pub fn calculate_available_rewards_for_pool(
+		liquidity_asset_id: TokenId,
+		block_number: u32,
+	) -> Result<Balance, DispatchError> {
+		let mut already_claimed_pool: Balance = 0;
+		if LiquidityMiningPoolClaimed::<T>::contains_key(&liquidity_asset_id) {
+			already_claimed_pool =
+				Self::liquidity_mining_pool_claimed(&liquidity_asset_id).unwrap();
```
mateuszaaa, line 566: again if you change the storage type to non-Option you wouldnt have to unwrap here ;)

pallets/xyk/src/lib.rs
```diff
@@ -512,10 +535,366 @@ pub mod pallet {
...
+
+		Ok(user_mangata_rewards_amount)
+	}
+
+	pub fn calculate_rewards_amount(
+		user: AccountIdOf<T>,
+		liquidity_asset_id: TokenId,
+		block_number: u32,
+	) -> Result<(Balance, i128), DispatchError> {
+		let current_time = block_number / 1000;
```
mateuszaaa, line 610: please avoid magic numbers and use  named  const variable instead of  `1000`

pallets/xyk/src/lib.rs
```diff
@@ -512,10 +535,366 @@ pub mod pallet {
...
+	}
+
+	pub fn calculate_work(
+		asymptote: Balance,
+		time: u32,
+		last_checkpoint: u32,
+		cummulative_work_in_last_checkpoint: U256,
+		missing_at_last_checkpoint: U256,
+	) -> Result<U256, DispatchError> {
+		// let (last_checkpoint,cummulative_work_in_last_checkpoint,missing_at_last_checkpoint,  )= UsersMinting::<T>::get((&user, &liquidity_asset_id)).unwrap();
```
mateuszaaa, line 631: remove comment

pallets/xyk/src/lib.rs
```diff
@@ -512,10 +535,366 @@ pub mod pallet {
...
+
+		let pool_work = Self::calculate_work(
+			liquidity_assets_amount,
+			current_time,
+			last_checkpoint,
+			cummulative_work_in_last_checkpoint,
+			missing_at_last_checkpoint,
+		)?;
+
+		Ok(pool_work)
```
mateuszaaa, line 676:
```
Self::calculate_work(
			liquidity_assets_amount,
			current_time,
			last_checkpoint,
			cummulative_work_in_last_checkpoint,
			missing_at_last_checkpoint,
		)
```
should work, return types are compatible

mateuszaaa (changes requested):
few minor comments

iStrike7 (commented):
Why are we using i128 in some places?
The libm library we are using, how are its performance characteristics? Does it solve our problems?
In some places we are using U256 containers to multiply more than 2 numbers, is that safe? Are the computations resilient to saturation?

stanieltron:
> Why are we using i128 in some places? The libm library we are using, how are its performance characteristics? Does it solve our problems? In some places we are using U256 containers to multiply more than 2 numbers, is that safe? Are the computations resilient to saturation?

-i128 is used because this numbers can go to negative values (wanted in this implementation)
-libm, not sure about performance, prob will be reworked for different calculation, where libm will not be necessary
-U256 is used to prevent overflow, not yet tested if it is enough for every case

mateuszaaa (approved):
lgtm

PR #179, opened 2022-01-20
Configs for public testnet and kusama mainnet added in chainspec

iStrike7:
Nothing about the current config was modified expect for token properties, protocol ids and protocol names.
Two new chainspecs were introduced public-testnet and kusama-mainnet.
The kusama-mainnet config has placeholding bridge AppIds, MGA ERC20 token contract address, and relay_chain and para_id  parameters.

No api changes.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -15,6 +15,32 @@ use sp_runtime::{
 	Perbill,
 };
 
+pub mod public_testnet_keys {
+	pub const ALICE_SR25519: &str =
+		"0x76e810a1f116b779fea0962f4102e9acb3d22ba603999ecb08ad163582de192c";
+	pub const BOB_SR25519: &str =
+		"0x2eec12f5af27c95fd5661323e544dc857155ea50689eaccdfd580af7c1be921c";
+	pub const CHARLIE_SR25519: &str =
+		"0xbc443b4f7023de2d868f74f9e51159961482dc46f76aa90a1c6ce58efff4be6a";
+	pub const SUDO_SR25519: &str =
+		"0x249af97c2ab99f229cdf18cc966833b894ae7c4d94c13fa86341209c64c8ec18";
+	pub const RELAY_SR25519: &str =
+		"0x7481b06f37b3500bb6ec8d569d2cede4ffcb151daee75f7de20c5bda2e22bb13";
+}
+
+pub mod kusama_mainnet_keys {
+	pub const ALICE_SR25519: &str =
+		"0x02d3074216e37c4e96c3a35be89fa27b6022fe08f02051989ed5b94768e69652";
+	pub const BOB_SR25519: &str =
+		"0xac1d5ec7cf53260c5ea1bb6be0d4fd8b23c50c088fad593de7cb60f76de4fe21";
+	pub const CHARLIE_SR25519: &str =
+		"0x708dbfb26bdf220b53443ff823da3e28845ddbd0d0aab1babd6074ac99b7b254";
+	pub const SUDO_SR25519: &str =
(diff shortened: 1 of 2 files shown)


node/src/chain_spec.rs
```diff
@@ -88,21 +114,353 @@ pub fn mangata_inflation_config() -> InflationInfo<Balance> {
...
+	properties.insert("tokenSymbol".into(), "MGX".into());
+	properties.insert("tokenDecimals".into(), 18.into());
+	properties.insert("ss58Format".into(), 42.into());
+
+	ChainSpec::from_genesis(
+		// Name
+		"Mangata Kusama Mainnet",
+		// ID
+		"mangata_kusama_mainnet",
+		ChainType::Development,
```
mateuszaaa, line 129: shouldnt that be `Local` or `Live` ? 
https://github.com/mangata-finance/substrate/blob/mangata-dev/primitives/chain-spec/src/lib.rs#L26-L33

node/src/chain_spec.rs
```diff
@@ -88,21 +114,353 @@ pub fn mangata_inflation_config() -> InflationInfo<Balance> {
...
+					),
+					(
+						kusama_mainnet_keys::BOB_SR25519.parse::<sr25519::Public>().unwrap().into(),
+						kusama_mainnet_keys::BOB_SR25519.parse::<sr25519::Public>().unwrap().into(),
+					),
+				],
+				// Initial relay account
+				kusama_mainnet_keys::RELAY_SR25519.parse::<sr25519::Public>().unwrap().into(),
+				// Sudo account
+				kusama_mainnet_keys::SUDO_SR25519.parse::<sr25519::Public>().unwrap().into(),
```
mateuszaaa, line 152: // offtop, 

do we want sudo account on kusama ? who will be in charge such account? Doesnt it imply that someone could mint MGA ?

node/src/chain_spec.rs
```diff
@@ -88,21 +114,353 @@ pub fn mangata_inflation_config() -> InflationInfo<Balance> {
...
+	properties.insert("tokenSymbol".into(), "MGAT".into());
+	properties.insert("tokenDecimals".into(), 18.into());
+	properties.insert("ss58Format".into(), 42.into());
+
+	ChainSpec::from_genesis(
+		// Name
+		"Mangata Public Testnet",
+		// ID
+		"mangata_public_testnet",
+		ChainType::Development,
```
mateuszaaa, line 295: Local?

mateuszaaa (changes requested):
Few minor questions. Also you should sync up with @goncer because you changed the unit names and its affects e2e-tests

https://github.com/mangata-finance/mangata-node/runs/4885933288?check_suite_focus=true#step:18:1285

node/src/chain_spec.rs
gleb-urvanov: It should be Live

node/src/chain_spec.rs
gleb-urvanov: Yes, right now we do want to have it, and with one of the first upgrades we will get rid of it.

node/src/chain_spec.rs
gleb-urvanov: Live

gleb-urvanov (changes requested):
Please alter the chain type

node/src/chain_spec.rs
iStrike7: Right! Changed.
Thanks for noticing that.

node/src/chain_spec.rs
iStrike7: Changed.

gleb-urvanov (approved):

mateuszaaa (approved):
LGTM

PR #184, opened 2022-01-25
skip publish if no build

goncer:
Fix the case for manual run skipping build  \-> hence no publish.

https://github.com/mangata\-finance/mangata\-node/actions/runs/1740760803


mateuszaaa (approved):
missed that in the first place!

PR #193, opened 2022-02-10
Feature/issuance vesting crowdloan

iStrike7:
This PR should be merged along with the PR of a similarly named branch in the moonbeam, substrate, orml, and crowdloan-rewards repos.

There are no API changes as such, except for a few removed extrinsics, modified storage and constants for staking.
A new pallet, pallet-issuance has been introduced. That is called by staking upon the start of a new session to compute the new sessions issuance. This pallet should be used to determine the issuance in a given session.

A fix was applied to the staking start_session functionality, the rounds and sessions are now in sync.

Initial tokens distribution has been changed, and also the characteristics of inflation.
The proceeds from transaction fees and tips now go entirely to the collator and none to treasury.

Please refer to below resources for further details:
https://github.com/mangata-finance/wiki/wiki/Mangata-participants-incentivization
https://github.com/mangata-finance/wiki/wiki/Mangata-fee-structure

The substrate vesting pallet is integrated and configured. The vesting tokens are issued by orml-tokens at genesis (directly via chainspec), issuance genesis config also accounts for it when calculating tge.

The vesting starts at block 0, and the vesting period is set to blocks equivalent to 1 year.

The pallet had to be cloned as pallet-vesting-mangata to avoid conflicts with polkadot, that also uses it. The pallet was modified for a different genesis input config and the extrinsic vested_transfer was removed.

The orml-tokens pallet was also modified to work with the vesting genesis issuance.

The crowdloan-rewards pallet from pureStake was modified and integrated.
It was modified to use a certain amount allocated for the rewards which the pallet mints upon init via extrinsic and when the user claims over time. The origin implementation in contrast was configured to mint the entire allocation at once and then transfer the amount over time.

Since we will be using mga chain blocks and not relay blocks as sense of time, all cumulus references were removed.
A few minor changes were made for safety.

The issuance pallet does not mint the crowdloan rewards at all. It does however, account for them when calculating the linear_issuance per session.


runtime/src/lib.rs
```diff
@@ -487,7 +487,7 @@ type ORMLCurrencyAdapterNegativeImbalance = <orml_tokens::CurrencyAdapter::<Runt
 pub struct ToAuthor;
 impl OnUnbalanced<ORMLCurrencyAdapterNegativeImbalance> for ToAuthor {
 	fn on_nonzero_unbalanced(amount: ORMLCurrencyAdapterNegativeImbalance) {
-		<orml_tokens::CurrencyAdapter::<Runtime, MgaTokenId> as PalletCurrency<AccountId>>::resolve_creating(&Authorship::author(), amount);
+		<orml_tokens::CurrencyAdapter::<Runtime, MgaTokenId> as PalletCurrency<AccountId>>::resolve_creating(&Authorship::author().unwrap(), amount);
```
iStrike7, line 490: This unwrap is dangerous, no?

iStrike7: Maybe check if `Authorship::author()` exists, if not then, just skip... This will result into the amount being burnt...

mateuszaaa (approved):
LGTM

PR #195, opened 2022-02-14
Feature/council as staking fallback

iStrike7:
To be merged after https://github.com/mangata-finance/mangata-node/pull/193

To be merged with similarly named PRs in the moonbeam and substrate repos

This PR implements the council as a fallback set if the staking yields no candidates, if the council is also empty then None is returned, in which case the session pallet will use the previous set.

No API changes


mateuszaaa (approved):
LGTM

PR #212, opened 2022-03-16
Audit/issue2

stanieltron:
addressed potential integer underflow with replacing sub with checked_sub

### DIFF
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -341,6 +341,7 @@ pub mod pallet {
 		/// Liquidity token cretion failed
 		LiquidityTokenCreationFailed,
 		NotEnoughtRewardsEarned,
+		SoldAmountTooLow,
 	}
 
 	#[pallet::event]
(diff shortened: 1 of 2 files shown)


iStrike7 (approved):

pallets/xyk/src/lib.rs
```diff
@@ -1614,8 +1615,10 @@ impl<T: Config> XykFunctionsTrait<T::AccountId> for Pallet<T> {
 				sold_asset_id.into(),
 				&sender,
 				&vault,
-				(sold_asset_amount - buy_and_burn_amount - treasury_amount - pool_fee_amount)
-					.into(),
+				(sold_asset_amount
+					.checked_sub(buy_and_burn_amount + treasury_amount + pool_fee_amount)
```
mateuszaaa, line 1619: what about possible overflow with `+`. Shouldt we use checked_add everywhere ?

pallets/xyk/src/lib.rs
```diff
@@ -1784,8 +1787,10 @@ impl<T: Config> XykFunctionsTrait<T::AccountId> for Pallet<T> {
...
```
mateuszaaa, line 1791: checked add ?

mateuszaaa (approved):
overall LGTM ,but few comments just to be sure

pallets/xyk/src/lib.rs
stanieltron: buy_and_burn_amount and the rest are 0.05% of sold amount which is capped u128, so no need

PR #224, opened 2022-04-04
issuance pallet pool_promotion api call added

stanieltron:
added api call to issuance pallet from xyk promote_pool fn


mateuszaaa (approved):
lgtm

PR #229, opened 2022-04-13
pallet bootstrap

mateuszaaa:
benchmarks stil missing


iStrike7 (changes requested):
Genesis config wont work.
Also *numerator

Haven't reviewed it deeply though

goncer:
@stanieltron  maybe you can take a look while @mateuszaaa  is on holidays? would  be awesome have it working so  can unblock testing + development activities

iStrike7 (approved):

PR #236, opened 2022-04-26
Feature/add node js and e2e repo for monitoring

goncer:


goncer:
Agreed that wont be the right approach. 
Closing.

PR #247, opened 2022-05-09
docker refactor

mateuszaaa:
- srtool alignment
- small progress
- progressing
- working compilation using docker compose
- update docker
- docker-cargo script
- progress'
- fix docker image builder script
- final state
- reduce images size & use use cargo cache
- fix for hosts without cargo
- very simple workflow


.github/workflows/ci.yml
```diff
@@ -47,74 +48,83 @@ jobs:
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Pull builder image
+        run: |
+          docker pull ${{ env.DOCKER_BUILDER_IMAGE }}
+      -
+        name: Run clippy
+        run: |
+          docker pull mangatasolutions/node-builder:0.1
+          ./docker-cargo.sh fetch || ./docker-cargo.sh fetch || ./docker-cargo.sh fetch || ./docker-cargo.sh fetch || ./docker-cargo.sh fetch
```
goncer, line 88: why this? :O

goncer (approved):
awesome!

.github/workflows/ci.yml
mateuszaaa: https://stackoverflow.com/questions/68283463/cargo-fails-to-download-dependencies-when-running-in-github-actions

was struggling with that for several days and couldn't figure out the solution. the cargo we have is even newer than mentioned in above post but problem still occasionally exists

PR #250, opened 2022-05-13
created a fast-setup kusama build, published and bumped the version to .21 for relay

goncer:


goncer:
I published a new version of kusama relay node that was built with fast runtime.

docker build -f ./scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile --build-arg BUILD_FLAGS=--features=fast-runtime  . -t polkadot_fast_runtime_921

PR #304, opened 2022-08-09
[onchain-fees] replace KAR token with TUR for on chain fees

vanderian:
KAR token is not needed currently
use the TUR token instead we need for bootstrap


mateuszaaa (approved):
LGTM :+1: so we want to do bootstrap on rococo ?

vanderian:
> LGTM +1 so we want to do bootstrap on rococo ?

no, just initial testing

PR #305, opened 2022-08-09
[xcm] add impl to kusama

vanderian:
add xcm config to kusama
depends on TUR token PR: https://github.com/mangata-finance/mangata-node/pull/304


vanderian:
copy&paste

iStrike7 (approved):
Couldn't find any typos

PR #312, opened 2022-08-16
[proxy] add pallet for proxy

vanderian:


runtime/mangata-kusama/src/lib.rs
```diff
@@ -1077,6 +1079,82 @@ impl xcm_asset_registry::Config for Runtime {
...
+			ProxyType::CancelProxy =>
+				matches!(c, Call::Proxy(pallet_proxy::Call::reject_announcement { .. })),
+			ProxyType::AutoCompound => {
+				matches!(
+					c,
+					Call::Xyk(pallet_xyk::Call::mint_liquidity { .. }) |
+						// Call::Xyk(pallet_xyk::Call::claim_rewards_all { .. })
+						Call::Xyk(pallet_xyk::Call::claim_rewards { .. })
+				)
+			},
```
vanderian, line 1130: to be defined later when we have a complete set of extrinsics working for auto compound
use this just as an example for now

PR #314, opened 2022-08-21
Automatic liq tokens activation on 'claim_liquidity_tokens' call

mateuszaaa:
- allow user to automatically activate liquidity tokens when claiming
- align benchmarks & weights
- rename claim_rewards -> claim_liquidity_tokens


iStrike7 (approved):
Looks alright

PR #318, opened 2022-08-25
activate bootstrap liquidity tokens using XYK pallet

mateuszaaa:
- create is_promote_pool helper fn
- create RewardsApi used for promoting liquidity tokens in bootstrap
- Create dedicated Api for Pool promotion and test edge cases
- implement Bootstrap::RewardsApi for Xyk
- remove duplicated benchmark
- allow pool promotion in bootstrap benchmarks


iStrike7 (approved):

pallets/bootstrap/src/tests.rs
```diff
@@ -1696,4 +1696,182 @@ fn archive_previous_bootstrap_schedules() {
...
+		// ACT
+		provisions(vec![
+			(PROVISION_USER1_ID, MGAId::get(), mga_provision, ProvisionKind::Regular),
+			(PROVISION_USER2_ID, KSMId::get(), ksm_provision, ProvisionKind::Vested(1, 150)),
+		]);
+
+		Bootstrap::on_initialize(100_u32.into());
+		assert_eq!(BootstrapPhase::Finished, Phase::<Test>::get());
+
+		Bootstrap::claim_and_activate_liquidity_tokens(Origin::signed(PROVISION_USER2_ID)).unwrap();
```
goncer, line 1767: @mateuszaaa  should not this fail with "not enough assets" ? or are you using MPL to reserve and then activate? 
or just hiding the error?

pallets/bootstrap/src/tests.rs
```diff
@@ -1696,4 +1696,182 @@ fn archive_previous_bootstrap_schedules() {
...
+		provisions(vec![
+			(PROVISION_USER1_ID, MGAId::get(), mga_provision, ProvisionKind::Regular),
+			(PROVISION_USER2_ID, KSMId::get(), ksm_provision / 2, ProvisionKind::Regular),
+			(PROVISION_USER2_ID, KSMId::get(), ksm_provision / 2, ProvisionKind::Vested(1, 150)),
+		]);
+
+		Bootstrap::on_initialize(100_u32.into());
+		assert_eq!(BootstrapPhase::Finished, Phase::<Test>::get());
+
+		Bootstrap::claim_and_activate_liquidity_tokens(Origin::signed(PROVISION_USER2_ID)).unwrap();
```
goncer, line 1807: is not the same as 
`test_activate_liq_tokens_is_called_with_all_liq_tokens_when_pool_is_promoted_and_all_provisions_are_non_vested`
?

pallets/bootstrap/src/tests.rs
```diff
@@ -1696,4 +1696,182 @@ fn archive_previous_bootstrap_schedules() {
...
+		provisions(vec![
+			(PROVISION_USER1_ID, MGAId::get(), mga_provision, ProvisionKind::Regular),
+			(PROVISION_USER2_ID, KSMId::get(), ksm_provision, ProvisionKind::Regular),
+		]);
+
+		Bootstrap::on_initialize(100_u32.into());
+		assert_eq!(BootstrapPhase::Finished, Phase::<Test>::get());
+
+		Bootstrap::claim_and_activate_liquidity_tokens(Origin::signed(PROVISION_USER1_ID)).unwrap();
+		Bootstrap::claim_and_activate_liquidity_tokens(Origin::signed(PROVISION_USER2_ID)).unwrap();
```
goncer, line 1839: this is a silent fail, right?

pallets/bootstrap/src/tests.rs
mateuszaaa: mpl is used underneath

pallets/bootstrap/src/tests.rs
mateuszaaa: nope, here provisions are it checks if only non vested part of provisions will be activated

in the other test user does only non vested provisions

pallets/bootstrap/src/tests.rs
mateuszaaa: that test makes sure that when pool is not promoted there `activate_liquidity` function is not called at all

PR #321, opened 2022-09-07
Develop v0.9.26

devdanco:


iStrike7 (commented):
Why isn't the change in scripts/run_benchmarks.sh showing up in git PR diff?

iStrike7 (commented):
I do not see RuntimeVersion changes in rococo or kusama either

iStrike7 (commented):
orml_asset_registry still seems to be named xcm_asset_registry

iStrike7 (commented):
generate_genesis_block(&spec, state_version) seems to be changed to generate_genesis_block(&*config.chain_spec, state_version), any particular reason why?

iStrike7 (approved):

PR #322, opened 2022-09-07
Fix job to run all the sequential instead of only critical

goncer:


majusko (approved):

goncer:
its easier to modify it once evth is merged

PR #337, opened 2022-10-20
Develop v0.9.29

devdanco:


iStrike7 (commented):
Need to look into bridge removal

PR #378, opened 2022-12-14
Feature [MGX-251] - Default transactional

iStrike7:
To be merged with https://github.com/mangata-finance/substrate/pull/75


pallets/xyk/src/tests.rs
```diff
@@ -1198,6 +1198,7 @@ fn sell_N_not_enough_selling_assset() {
 }
 
 #[test]
+#[ignore]
```
mateuszaaa, line 1201: cant we align those tests instead of disabling them

mateuszaaa (changes requested):
could you provide some UT that tests new impl ?

pallets/xyk/src/lib.rs
```diff
@@ -2303,18 +2314,36 @@ impl<T: Config> XykFunctionsTrait<T::AccountId> for Pallet<T> {
...
+					bought_asset_amount,
+					min_amount_out,
+				));
+			}
 		}
 
 		Ok(())
 	}
 
+	// To put it comprehensively the only reason that the user should lose out on swap fee
```
majusko, line 2334: Not sure if achievable, but would it be possible to align this comment with autogenerated RustDoc format we are currently using? cc: @mateuszaaa

runtime/mangata-kusama/src/lib.rs
```diff
@@ -160,11 +160,11 @@ impl_opaque_keys! {
 pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("mangata-parachain"),
 	impl_name: create_runtime_str!("mangata-parachain"),
-	authoring_version: 10,
-	spec_version: 10,
+	authoring_version: 11,
```
majusko, line 163: This should be incremented only when preparing the RC, so we don't have conflicts.

majusko (changes requested):

runtime/mangata-kusama/src/lib.rs
iStrike7: We need to do this, so that it doesn't fail ci.

majusko: Understand. Didn't know. Noting it down for some future CI improvements.

pallets/xyk/src/lib.rs
majusko: Ok, @mateuszaaa explained that it would be there like this. So ignore my comment.

mateuszaaa (approved):
lgtm

PR #400, opened 2023-01-13
[mgx-312][mgx-302] Feature/proxy audit

vanderian:
remove Any
change proxy reserve amounts

### DIFF
--- a/runtime/integration-test/src/proxy.rs
+++ b/runtime/integration-test/src/proxy.rs
@@ -9,7 +9,7 @@ const LP_ASSET_ID: u32 = 2;
 fn proxy_permissions_correct() {
 	ExtBuilder {
 		balances: vec![
-			(AccountId::from(ALICE), NATIVE_ASSET_ID, 1000 * UNIT),
+			(AccountId::from(ALICE), NATIVE_ASSET_ID, 10000 * UNIT),
 			(AccountId::from(BOB), NATIVE_ASSET_ID, 1000 * UNIT),
 			(AccountId::from(BOB), ASSET_ID_1, 1000 * UNIT),
 		],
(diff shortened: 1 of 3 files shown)


runtime/mangata-kusama/src/lib.rs
```diff
@@ -190,7 +190,7 @@ mod currency {
 	pub const DOLLARS: Balance = super::UNIT;
 
 	pub const fn deposit(items: u32, bytes: u32) -> Balance {
-		items as Balance * 15 * CENTS + (bytes as Balance) * 6 * CENTS
+		items as Balance * 5000 * DOLLARS + (bytes as Balance) * 60 * CENTS
```
vanderian, line 193: roughly 1000x multiple of
```	pub fn base_tx_in_mgr() -> Balance {
		UNIT
	}
``` 

for reference Polkadot:
```items as Balance * 20 * DOLLARS + (bytes as Balance) * 100 * MILLICENTS```
&
base_tx_in_ksm == CENT / 10

File: runtime/integration-test/src/proxy.rs [hunk 2]
```diff
@@ -5,73 +5,6 @@ type SystemError = frame_system::Error<Runtime>;
 const ASSET_ID_1: u32 = 1;
 const LP_ASSET_ID: u32 = 2;
 
-#[test]
```
mateuszaaa, line 8: shouldnt we rather align those ?

vanderian: we already have the same cases covered by the other test:
add proxy
proxy OK, inner call executed when extrinsic is allowed for proxy type
proxy OK, inner call filtered out when extrinsic is not allowed for proxy type
remove proxy
proxy FAIL with no proxied account

iStrike7 (approved):

PR #427, opened 2023-02-15
[identity][mgx-362] add pallet & integration test

vanderian:


runtime/integration-test/src/identity.rs
```diff
@@ -0,0 +1,34 @@
...
+			),
+			BadOrigin
+		);
+		// origin passes, but fails on no identity set
+		assert_noop!(
+			Identity::kill_identity(RuntimeOrigin::root(), AccountId::from(BOB).into()),
+			pallet_identity::Error::<Runtime>::NotNamed
+		);
+	});
+}
```
vanderian, line 34: test the integration for
```
type IdentityForceOrigin = EnsureRoot<AccountId>;
type IdentityRegistrarOrigin = EnsureRoot<AccountId>;
```

runtime/mangata-kusama/src/lib.rs
```diff
@@ -1655,13 +1684,13 @@ construct_runtime!(
 		Sudo: pallet_sudo_mangata::{Pallet, Call, Config<T>, Storage, Event<T>} = 49,
 		SudoOrigin: pallet_sudo_origin::{Pallet, Call, Event<T>} = 50,
 		Council: pallet_collective_mangata::<Instance1>::{Pallet, Call, Storage, Origin<T>, Event<T>, Config<T>} = 51,
+		Identity: pallet_identity::{Pallet, Call, Storage, Event<T>} = 56,
```
vanderian, line 1687: is there some convention which number to use ?

majusko (dismissed):

runtime/mangata-kusama/src/lib.rs
mateuszaaa: i guess you can consider updating pallets Id, the only thing you need on top of that seems to be incrementation of transaction_version of runtime version struct, but it has been done already as thats something we do with every release :

https://github.com/mangata-finance/substrate/blob/8d5b532ef66f0f06be2e9b23c0f8ecf335440075/primitives/version/src/lib.rs#L201-L202

https://github.com/mangata-finance/mangata-node/blob/56c044d97da917b005ec7dfa696f657c3917e557/runtime/mangata-rococo/src/lib.rs#L181

runtime/mangata-rococo/src/lib.rs
```diff
@@ -1559,6 +1559,36 @@ impl pallet_proxy::Config for Runtime {
...
+	type FieldDeposit = FieldDeposit;
+	type SubAccountDeposit = SubAccountDeposit;
+	type MaxSubAccounts = MaxSubAccounts;
+	type MaxAdditionalFields = MaxAdditionalFields;
+	type MaxRegistrars = MaxRegistrars;
+	type ForceOrigin = IdentityForceOrigin;
+	type RegistrarOrigin = IdentityRegistrarOrigin;
+	type Slashed = Treasury;
+	type WeightInfo = pallet_identity::weights::SubstrateWeight<Runtime>;
+}
```
mateuszaaa, line 1590: dont we need to do the same for kusama runtime ?

runtime/mangata-rococo/src/lib.rs
```diff
@@ -1559,6 +1559,36 @@ impl pallet_proxy::Config for Runtime {
...
+parameter_types! {
+	// Add one item in storage and take 258 bytes
+	pub const BasicDeposit: Balance = deposit(1, 256);
+	// No item in storage but takes 66 bytes
+	pub const FieldDeposit: Balance = deposit(0, 66);
+	// No item in storage but takes 53 bytes
+	pub const SubAccountDeposit: Balance = deposit(0, 53);
+	pub const MaxSubAccounts: u32 = 100;
+	pub const MaxAdditionalFields: u32 = 100;
+	pub const MaxRegistrars: u32 = 20;
```
mateuszaaa, line 1571: Is it maximum numbers of collators identities that can be set, or i got it wrong ?

runtime/mangata-rococo/src/lib.rs
```diff
@@ -1559,6 +1559,36 @@ impl pallet_proxy::Config for Runtime {
...
 
+parameter_types! {
+	// Add one item in storage and take 258 bytes
+	pub const BasicDeposit: Balance = deposit(1, 256);
+	// No item in storage but takes 66 bytes
+	pub const FieldDeposit: Balance = deposit(0, 66);
+	// No item in storage but takes 53 bytes
+	pub const SubAccountDeposit: Balance = deposit(0, 53);
+	pub const MaxSubAccounts: u32 = 100;
+	pub const MaxAdditionalFields: u32 = 100;
```
mateuszaaa, line 1570: if you could provide some docs what these params are about that would be very helpful for the future

mateuszaaa (changes requested):

runtime/mangata-rococo/src/lib.rs
vanderian: all of these are pretty nicely documented in the identity pallet, i think that should suffic
i can add some comments for the specific values thou

runtime/mangata-rococo/src/lib.rs
vanderian: a registrar is an entity which can pass a judgement on identity
an identity verifier of sorts
this number defines max amount of such registrars

user sets an identity -> a registrar can verify the fields to set some confidence level on the submitted fields

runtime/mangata-rococo/src/lib.rs
vanderian: applied to both runtimes

runtime/mangata-kusama/src/lib.rs
```diff
@@ -1629,6 +1629,35 @@ impl pallet_proxy::Config for Runtime {
 	type AnnouncementDepositBase = AnnouncementDepositBase;
 	type AnnouncementDepositFactor = AnnouncementDepositFactor;
 }
+parameter_types! {
+	// Add item in storage and take 270 bytes, Registry { [], Balance, Info { [], [u8,32] * 7, [u8,20] }}
+	pub const BasicDeposit: Balance = deposit(1, 270);
```
vanderian, line 1634: eg. bytes { 1, 16,  { 1, 33 * 7, 21 } = 270

runtime/mangata-kusama/src/lib.rs
```diff
@@ -1670,6 +1699,9 @@ construct_runtime!(
 		} = 1,
 		Timestamp: pallet_timestamp::{Pallet, Call, Storage, Inherent} = 2,
 		ParachainInfo: parachain_info::{Pallet, Storage, Config} = 3,
+		Utility: pallet_utility_mangata::{Pallet, Call, Event} = 4,
+		Proxy: pallet_proxy::{Pallet, Call, Storage, Event<T>} = 5,
+		Maintenance: pallet_maintenance::{Pallet, Call, Storage, Event<T>} = 6,
 
 		// Monetary stuff.
```
vanderian, line 1706: xyk needs to stay unchanged due to OAK's XCM transacts for autocompound

monetary ids = 10->29

mateuszaaa (approved):
lgtm

PR #438, opened 2023-03-06
[xcm][mgx-361] Feature/xcm refactor

vanderian:


runtime/mangata-kusama/src/xcm_config.rs
```diff
@@ -291,7 +193,7 @@ pub type XcmRouter = (
 
 impl pallet_xcm::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
-	type SendXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
+	type SendXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, ()>;
```
vanderian, line 196: similar to statemine, i though it would be useful to also disable arbitrary xcm calls
even though we have them filtered in XcmExecuteFilter

same change for rococo

runtime/integration-test/src/setup.rs
```diff
@@ -38,6 +59,21 @@ impl Default for ExtBuilder {
 }
 
 impl ExtBuilder {
+	pub fn assets(mut self, assets: Vec<(TokenId, AssetMetadataOf)>) -> Self {
+		self.assets = assets;
+		self
+	}
+
+	pub fn balances(mut self, balances: Vec<(AccountId, TokenId, Balance)>) -> Self {
```
mateuszaaa, line 67: :heart:

runtime/integration-test/src/lib.rs
```diff
@@ -11,3 +11,6 @@ mod bootstrap;
 
 #[cfg(any(feature = "with-kusama-runtime",))]
 mod proxy;
+
+#[cfg(any(feature = "with-kusama-runtime",))]
```
mateuszaaa, line 15: recently we switched to building both envs at once (kusama & rococo) so maybe we can get rid of conditional compilation ? :thinking:

runtime/integration-test/src/xcm/kusama_xcm_transfer.rs
```diff
@@ -0,0 +1,210 @@
...
+		));
+	});
+
+	KusamaRelay::execute_with(|| {
+		assert_eq!(kusama_runtime::Balances::free_balance(&AccountId::from(BOB)), unit(12) - fee);
+	});
+}
+
+#[test]
+fn transfer_asset() {
```
mateuszaaa, line 83: are these tests run on ci ?

mateuszaaa (approved):
lgtm

runtime/integration-test/src/lib.rs
vanderian: this is different, the feature flag is used to define which runtime to use in the tests
it is because of the imports mostly
in setup we use `with-some-runtime` to import from specific runtimes
or the fees use a bit different values when it comes to rococo vs kusama

we only do tests for kusama now, but we also should for rococo, and eventually polkadot

runtime/integration-test/src/xcm/kusama_xcm_transfer.rs
vanderian: they were run as a part of `cargo test` previously, will check

vanderian: verifed, they are run as part of test

PR #464, opened 2023-03-31
Bootstrap finalize upgrade

iStrike7:


pallets/bootstrap/src/lib.rs
```diff
@@ -681,38 +681,80 @@ pub mod pallet {
...
+			);
+
+			ensure!(
+				VestedProvisions::<T>::iter_keys().next().is_none(),
+				Error::<T>::BootstrapMustBePreFinalized
+			);
+
+			ensure!(
+				WhitelistedAccount::<T>::iter_keys().next().is_none(),
+				Error::<T>::BootstrapMustBePreFinalized
```
majusko, line 745: Can we have a specific error message for each error? It would be hard to debug when all events are the same.

majusko (changes requested):

pallets/bootstrap/src/lib.rs
```diff
@@ -659,20 +661,18 @@ pub mod pallet {
...
 		/// bootstrap might not fit in a single block. As a result tx can be rejected !!!**
 		#[pallet::call_index(7)]
-		#[pallet::weight(<<T as Config>::WeightInfo>::finalize().saturating_add(T::DbWeight::get().reads_writes(1, 1).saturating_mul(Into::<u64>::into(*limit).saturating_add(u64::one()))))]
+		#[pallet::weight(Weight::from_ref_time(40_000_000)
+							.saturating_add(T::DbWeight::get().reads_writes(6, 0)
+								.saturating_add(T::DbWeight::get().reads_writes(1, 1).saturating_mul(Into::<u64>::into(T::ClearStorageLimit::get())))))]
 		#[transactional]
-		pub fn finalize(origin: OriginFor<T>, mut limit: u32) -> DispatchResult {
-			ensure_root(origin)?;
+		pub fn pre_finalize(origin: OriginFor<T>) -> DispatchResult {
```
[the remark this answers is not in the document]
iStrike7: > If there an action which require iterative execution and it should be finished before the bootstrap can be finalized, it should be a separate extrinsic. https://mangatafinance.atlassian.net/browse/PRO-107?search_id=1277c0ab-6b46-4295-996f-5a326d251a23

pallets/bootstrap/src/lib.rs
iStrike7: Why? They all pertain to the same thing.

majusko: It's not clear which case failed from the event name. Either add the message, different event or log a warning.

iStrike7: That's the point it doesn't matter which case failed. If the BootstrapMustBePreFinalized error appears then no matter which "case failed", the bootstrap must be *indifferently* pre-finalized using the pre_finalize extrinsic. They are just in different ensures for readability and easily accounting reads.

PR #468, opened 2023-04-05
[mgx-361][xcm] add receive tests

vanderian:
add tests for receiving asset on mangata
* no location
* no fee


runtime/integration-test/src/xcm/kusama_xcm_transfer.rs
```diff
@@ -193,6 +191,139 @@ fn transfer_asset() {
...
+				.into()
+			),
+			WeightLimit::Limited(600_000_000),
+		));
+
+		assert_eq!(Tokens::free_balance(NATIVE_ASSET_ID, &AccountId::from(ALICE)), 80 * unit);
+	});
+
+	// no asset found -> no FPS -> too expensive
+	Mangata::execute_with(|| {
```
mateuszaaa, line 223: is it dependant on `Sibling::execute_with` ? Can we maybe split it into multiple smaller test cases ?

majusko (approved):

PR #495, opened 2023-05-19
[MGX-436] Fix multi swap prevalidation

mateuszaaa:
Prevalidate multiswaps with length = 2 same as atomic swaps


vanderian (approved):
lgtm

would be nice to have the FeeHelpers in a common runtime module instead

PR #505, opened 2023-05-31
Bugfix/liq tokens MGX-434

majusko:


pallets/xyk/src/lib.rs
```diff
@@ -1270,6 +1273,12 @@ impl<T: Config> Pallet<T> {
 		Ok(result)
 	}
 
+	pub fn get_liq_tokens_for_trading() -> Result<Vec<TokenId>, DispatchError> {
+		return LiquidityAssets::<T>::iter().values().collect()
+			.ok_or_else(|| Error::<T>::UnexpectedFailure.into())
```
mateuszaaa, line 1278: you can use below to avoid unnecescary function call
```
			.ok_or(Error::<T>::UnexpectedFailure)
```

pallets/xyk/src/lib.rs
```diff
@@ -2852,6 +2858,13 @@ impl<T: Config> XykFunctionsTrait<T::AccountId> for Pallet<T> {
 			ExistenceRequirement::KeepAlive,
 		)?;
 
+		let liquidity_assets_minted = multiply_by_rational_with_rounding(
+			first_asset_amount,
+			total_liquidity_assets,
+			first_asset_reserve,
+			Rounding::Down,
+		).ok_or(Error::<T>::UnexpectedFailure)?;
```
mateuszaaa, line 2866: it seems that in case of total_liquidity_issuance == 0 it will evalute to 0 :thinking: ?

mateuszaaa (changes requested):

mateuszaaa (changes requested):
consider addint UT for scenario where pool is fully drained and then someone comes to "recreate it"

pallets/xyk/src/lib.rs
```diff
@@ -2784,29 +2793,26 @@ impl<T: Config> XykFunctionsTrait<T::AccountId> for Pallet<T> {
...
-		.ok_or(Error::<T>::UnexpectedFailure)?;
+		// The pool is empty and we are basically creating a new pool and reusing the existing one
+		if !(first_asset_reserve.is_zero() && second_asset_reserve.is_zero()) && !total_liquidity_assets.is_zero() {
 
-		ensure!(
-			second_asset_amount <= expected_second_asset_amount,
-			Error::<T>::SecondAssetAmountExceededExpectations,
-		);
+			// Calculation of required second asset amount and received liquidity token amount
+			ensure!(!first_asset_reserve.is_zero(), Error::<T>::DivisionByZero);
```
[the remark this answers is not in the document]
majusko: True, but it was there before like this. Should I change it to something else since updating this part?

mateuszaaa (approved):

mateuszaaa (dismissed):
lgtm

pallets/xyk/src/tests.rs
```diff
@@ -904,7 +906,7 @@ fn multiswap_sell_bad_atomic_swap_charges_fee_W() {
 
 		assert_ok!(XykStorage::multiswap_sell_asset(
 			RuntimeOrigin::signed(TRADER_ID),
-			vec![1, 2, 3, 6, 5],
+			vec![1, 2, 3, 4, 5],
```
majusko, line 909: This was probably an old bug in our tests that went unnoticed until this added validation

mateuszaaa (dismissed):
lgtm

PR #509, opened 2023-06-09
[MGX-521]Runtime api for checking FeeLocks

mateuszaaa:
Dedicated api for FE team, so they dont have to reimplement feelock logic

### DIFF
--- a/runtime/mangata-kusama/src/lib.rs
+++ b/runtime/mangata-kusama/src/lib.rs
@@ -34,6 +34,7 @@ use orml_traits::{
 pub use pallet_sudo_mangata;
 use pallet_transaction_payment_mangata::{ConstFeeMultiplier, Multiplier, OnChargeTransaction};
 use pallet_vesting_mangata_rpc_runtime_api::VestingInfosWithLockedAt;
+use xyk_runtime_api::SwapKind;
 // Polkadot Imports
 pub use polkadot_runtime_common::BlockHashCount;
 use scale_info::TypeInfo;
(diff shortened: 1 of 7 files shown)


runtime/mangata-kusama/src/lib.rs
```diff
@@ -1989,6 +1991,84 @@ impl_runtime_apis! {
...
+					Some(true)
+				}
+			}
+		}
+
+		fn is_buy_asset_lock_free(
+			path: Vec<TokenId>,
+			input_amount: Balance,
+			) -> Option<bool>{
+			match (path.len(), pallet_fee_lock::FeeLockMetadata::<Runtime>::get()) {
```
majusko, line 2038: I think this conditioning should be inside the pallet so we don't have it duplicated. Can we make such a function in the FeeLock pallet?

majusko (changes requested):

runtime/mangata-kusama/src/lib.rs
mateuszaaa: actually at the moment xyk & feelock pallets are not aware of each other, its only in runtime where informations from both pallet are utilized. ideally it should be plased somewhere in common runtime module so both kusama & rococo could use it, maybe we should consider another look on that :thinking:

majusko: Aaah, I see. Yes totally agree. Created ticket for it (MGX-534)

majusko (approved):

vanderian (approved):

PR #539, opened 2023-07-25
Storage migration asset registry

devdanco:


vanderian (dismissed):

runtime/mangata-kusama/src/migration.rs
```diff
@@ -0,0 +1,67 @@
...
+
+		let mut weight: Weight = Weight::zero();
+
+		orml_asset_registry::Metadata::<Runtime>::translate(
+			|_key, mut old_meta: AssetMetadataOf| {
+				weight.saturating_accrue(
+					<Runtime as frame_system::Config>::DbWeight::get().reads_writes(1, 1),
+				);
+
+				let issuance = orml_tokens::Pallet::<Runtime>::total_issuance(_key);
```
mateuszaaa, line 31: its either prefixed with `_` and unused, or not prefixed ;)

runtime/mangata-kusama/src/migration.rs
```diff
@@ -0,0 +1,67 @@
...
+				let name = sp_std::str::from_utf8(&old_meta.name);
+				if issuance.is_zero() && name.map_or(false, |n| n.starts_with("Liquidity")) {
```
mateuszaaa, line 33: you should rather use XYK::LiquidityAssets to check if given asset is liquidity asset

runtime/mangata-kusama/src/migration.rs
```diff
@@ -0,0 +1,67 @@
...
+				}
+			},
+		);
+
+		weight
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
+		Ok(Vec::new())
```
mateuszaaa, line 48: please check if there are any assets that require migration, so we can make sure that this migration will start failing on ci after runtime upgrade and we will not forget to remove it ;)

runtime/mangata-kusama/src/migration.rs
```diff
@@ -0,0 +1,67 @@
...
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(_: Vec<u8>) -> Result<(), &'static str> {
+		info!(
+			target: "asset_registry",
+			"post_upgrade: checks"
+		);
+		orml_asset_registry::Metadata::<Runtime>::iter().for_each(|(token_id, meta)| {
+			let issuance = orml_tokens::Pallet::<Runtime>::total_issuance(token_id);
+			let name = sp_std::str::from_utf8(&meta.name);
+			if name.map_or(false, |n| n.starts_with("Liquidity")) {
```
mateuszaaa, line 60: please use XYK to check whether asset is liquidity token or not

runtime/mangata-rococo/src/migration.rs
```diff
@@ -0,0 +1,67 @@
...
+		storage_prefix, unhashed,
+	},
+	traits::OnRuntimeUpgrade,
+	StoragePrefixedMap, Twox64Concat,
+};
+use log::info;
+use sp_runtime::traits::Zero;
+use xcm::IntoVersion;
+
+pub struct AssetRegistryMigration;
```
mateuszaaa, line 15: [optional] ideally AssetRegistryMigration should be put in `common_runtime`. Then you will have no access to `Runtime` directly but you can make AssetRegistryMigration generic over Runtime with proper bounds

https://github.com/mangata-finance/mangata-node/blob/3a395521bf47d15d6eeb669c36df1e1c7a8d6e8a/runtime/common/src/lib.rs#L457-L468

mateuszaaa (changes requested):

mateuszaaa (approved):

PR #541, opened 2023-07-25
RPC: tradeable tokens

devdanco:


devdanco:
Only for testing purpose

PR #603, opened 2023-09-28
Bugfix/liq tokens mgx 434 v2

majusko:
Added more validations for other extrinsic and APIs that interact with the pool. Basic test with the error trigger.


mateuszaaa (approved):
lgtm

PR #630, opened 2023-10-19
CI: e2e tests job fix MGX-785

tenequm:
MGX-785


goncer (approved):
I do not remember why we set that, but here is my approval! :shrug:

PR #637, opened 2023-11-09
CI: Update Helm chart version for dev envs to fix issue with deprecated parameters MGX-808

tenequm:
MGX-808

### DIFF
--- a/.github/workflows/branch-develop.yml
+++ b/.github/workflows/branch-develop.yml
@@ -51,6 +51,7 @@ jobs:
     outputs:
       GLOBAL_VERSION: ${{ steps.set_vars.outputs.GLOBAL_VERSION }}
       GIT_BRANCH: ${{ steps.set_vars.outputs.GIT_BRANCH }}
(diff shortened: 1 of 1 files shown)


devops/helmfiles/values.yaml.gotmpl
```diff
@@ -4,7 +4,7 @@ environment: {{ requiredEnv "ENVIRONMENT" | quote }}
 image:
   tag: {{ requiredEnv "IMAGE_TAG" | quote }}
 customChainspecUrl: {{ .Values | get "nodeCustomChainspecUrl" "" }}
-chain: {{ .Values | get "nodeChain" "rococo" }}
+chain: {{ .Values | get "nodeChain" "mangata-rococo-local" }}
```
vanderian, line 7: i think this should be just `mangata-rococo` ?
`rococo` previously it was the public testnet

`mangata-rococo-local` is a dev net
`mangata-rococo` is the public testnet

vanderian (approved):

PR #660, opened 2023-12-18
Rolldown + sequencer staking + asset registry

iStrike7:


pallets/sequencer-staking/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+		Ok(().into())
+	}
+
+	fn process_potential_authors(authors: Vec<(AccountIdOf<T>, BalanceOf<T>)>) -> Option<Vec<(AccountIdOf<T>, BalanceOf<T>)>> {
+		let minimal_stake_amount = MinimalStakeAmount::<T>::get();
+		let filtered_authors: Vec<_> = authors.into_iter().filter( |a| <SequencerStake<T>>::get(&a.0) >= minimal_stake_amount).collect::<>();
+		if filtered_authors.len() as u32 >= T::MinimumSequencers::get(){
+			Some(filtered_authors)
+		} else {
+			None
```
goncer, line 167: What I do understand here is that for now the minimal sequencer stake is mandatory for any collator. But what I do understand from the doc, is that we should not filter  by this sequencer stake unless we have two fo them? 
Is not quite risky return no candidates? whill this halt the chain?

iStrike7: In the parachain-staking pallet you can see that if this function returns none - the filter is not applied.

goncer: niice!

PR #661, opened 2023-12-19
rolldown prototype

stanieltron:
all functionalities rolldown related including sequencer stake (excluding ferries)


pallets/rolldown/src/lib.rs
```diff
@@ -0,0 +1,916 @@
...
+
+					// Add a new variable of type Requests
+					let requests: Requests = deserialized;
+
+					if requests.lastProccessedRequestOnL1 <
+						Self::get_prev_last_processed_request_on_l2()
+					{
+						log!(info, "lastProccessedRequestOnL1 is less than prev_last_processed_request_on_l2");
+					} else {
+						for request_id in (requests.lastProccessedRequestOnL1 + 1_u128)..=
```
goncer, line 409: max number of requests?

pallets/rolldown/src/lib.rs
```diff
@@ -0,0 +1,916 @@
...
+		// no checks if this read is correct, can put counters out of sync
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::DbWeight::get().reads_writes(1, 1).saturating_add(Weight::from_parts(40_000_000, 0)))]
+		pub fn council_update_l2_from_l1(
+			origin: OriginFor<T>,
+			input_json: String,
+		) -> DispatchResultWithPostInfo {
+			let _ = ensure_root(origin)?;
+
+			match Self::deserialize_json(&input_json) {
```
goncer, line 397: will it be a max size of this json?

pallets/rolldown/src/lib.rs
```diff
@@ -0,0 +1,916 @@
...
+			// check json length to prevent big data spam, maybe not necessary as it will be checked later and slashed
+			let current_block_number = <frame_system::Pallet<T>>::block_number();
+			let dispute_period_end = current_block_number + (DISPUTE_PERIOD_LENGTH as u32).into();
+
+			let l1_pending_requests_json = input_json.trim().to_string();
+
+			// ensure sequencer has rights to update
+			if let Some(sequencer) = SEQUENCER_RIGHTS::<T>::get(&sequencer) {
+				if sequencer.readRights == 0 {
+					log!(info, "{:?} does not have sufficient readRights", sequencer);
```
goncer, line 273: could it happen that all the sequencers end up with zero reads ? what would happen if so?

pallets/rolldown/src/lib.rs
```diff
@@ -0,0 +1,916 @@
...
+		StorageMap<_, Blake2_128Concat, u128, Update<T::AccountId>, OptionQuery>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {}
+
+	#[pallet::error]
+	/// Errors
+	pub enum Error<T> {
+		OperationFailed,
```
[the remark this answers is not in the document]
stanieltron: sure

pallets/rolldown/src/lib.rs
stanieltron: not limited for now, not sure how to limit it from concept perspective

pallets/rolldown/src/lib.rs
stanieltron: this is same issue as max number of requests

pallets/rolldown/src/lib.rs
stanieltron: yes it could, there is an emergency root only fn, which will do council_update_l2_from_l1. Same as regular, but requests from this json are not going to the queue and cannot be canceled, they are executed right away. Amongst these request are multiple cancel resolutions (if there are no read rights, there must have been a lot of cancels). When cancel resolutions are processed, slashing occures and read rights are returned.

PR #684, opened 2024-01-31
feat: add eigen chainspec

devdanco:


mateuszaaa (approved):
lgtm

mateuszaaa:
it was created before we migrated to solochain, so no longer needed

mateuszaaa (approved):
lgtm

mateuszaaa (changes requested):
i think we should have new chainspec called `holesky` because in local testnet token addresses will not match

PR #700, opened 2024-02-26
[rolldown] council actions for rolldown pallet

mateuszaaa:
- council action for update & cancel requests
- L1UpdateBuilder helper
- test for invalid counters


iStrike7 (approved):
Approved without review

PR #704, opened 2024-02-29
feat: add rpc for verify pending requests

devdanco:


pallets/rolldown/runtime-api/src/lib.rs
```diff
@@ -1,10 +1,12 @@
 // Copyright (C) 2021 Mangata team
 #![cfg_attr(not(feature = "std"), no_std)]
 
+use sp_core::H256;
 use sp_std::vec::Vec;
 sp_api::decl_runtime_apis! {
 	pub trait RolldownRuntimeApi{
 		fn get_pending_updates_hash() -> sp_core::H256;
 		fn get_pending_updates() -> Vec<u8>;
+		fn verify_pending_requests(hash: H256, request_id: u128) -> bool;
```
mateuszaaa: should return Option<bool>
- Some(true) - means request is ok
- Some(false) - means rrequest is nok
- None - means request does not exists

runtime/mangata-kusama/src/lib.rs
```diff
@@ -865,6 +865,10 @@ impl_runtime_apis! {
 		fn get_pending_updates() -> Vec<u8> {
 			pallet_rolldown::Pallet::<Runtime>::l2_update_encoded()
 		}
+
+		fn verify_pending_requests(hash: sp_core::H256, request_id: u128) -> bool {
+			pallet_rolldown::Pallet::<Runtime>::verify_pending_requests(hash, request_id).is_some()
```
mateuszaaa: this is not correct, if verify pendinhg requests returns Some(false) the outcome should be false

pallets/rolldown/rpc/src/lib.rs
```diff
@@ -71,4 +80,21 @@ where
...
+
+	fn verify_pending_requests(
+		&self,
+		hash: H256,
+		request_id: u128,
+		at: Option<<Block as BlockT>::Hash>,
+	) -> RpcResult<bool> {
+		let api = self.client.runtime_api();
+		let at = at.unwrap_or(self.client.info().best_hash);
+		api.verify_pending_requests(at, hash, request_id).map_err(|e| {
```
mateuszaaa, line 92: once verify_pending_requests returns Option 

api.verify_pending_requests witll return Result<Opion<bool>>

first result is set to error when there is some problem with runtime api execution, in that case we should raise error
When  option is set to None we also should raise an error because that means requested request_id does not exists

mateuszaaa (changes requested):

mateuszaaa (approved):

PR #714, opened 2024-03-16
Eth rollup develop solo seq mod

iStrike7:


rollup/node/src/chain_spec.rs
```diff
@@ -293,12 +293,19 @@ fn rollup_genesis(
...
+			sequencers_stake: if initial_collators_as_sequencers {
+				initial_authorities
+					.iter()
+					.map(|(acc, _)| (acc.clone(), 10_000_000_u128))
+					.collect()
 			} else {
 				Default::default()
 			},
 		},
+		// Use sequncer staking genesis config - this is intended only for testing
```
mateuszaaa:
```suggestion
		// Use sequencer staking genesis config - this is intended only for testing
```

pallets/rolldown/src/mock.rs
```diff
@@ -184,6 +184,10 @@ impl ExtBuilder {
 				MockSequencerStakingProviderApi::is_active_sequencer_context();
 			is_liquidity_token_mock.expect().return_const(true);
 
+			let is_selected_sequencer_mock =
```
mateuszaaa, line 192: :+1:

rollup/node/src/chain_spec.rs
```diff
@@ -293,12 +293,19 @@ fn rollup_genesis(
...
+		rolldown: rollup_runtime::RolldownConfig { sequencers: vec![] },
```
mateuszaaa: how about just removing genesis config for rolldown.

then in test instead of initializing genesis config with inital sequencers, we could just call

```
				Pallet::<T>::new_sequencer_active(s);

```

instead?

mateuszaaa: https://github.com/mangata-finance/mangata-node/pull/714/files#diff-594088d0312e379b90937ff1434ab8ca1e2b1a444927cabb4a58de2104849051R163-R167

goncer (approved):

rollup/node/src/chain_spec.rs
iStrike7: This is what I was about to do when I realized that there's https://github.com/mangata-finance/mangata-node/blob/b5ab95aedf52dabbe292b0083dfdbba1bc7f8d7a/pallets/rolldown/src/lib.rs#L276
so we're gonna need a genesis config anyway (you could set that as the storage default value and get rid of that, but that might get a little confusing for someone reading the code)

PR #715, opened 2024-03-18
fix: update configs for fungible env deployments and remove `release` env configs for `eth-rollup-develop` branch related deployments MGX-944

tenequm:
MGX-944


mateuszaaa (approved):
lgtm

PR #718, opened 2024-03-21
[fix] consider withdrawal resolutions

mateuszaaa:
- consider withdrawal resolutions when validating sequencer update


tenequm (approved):
lgtm

PR #722, opened 2024-04-01
Added L1Asset info for ETH in chainspec

iStrike7:
To be merged with
https://github.com/mangata-finance/open-runtime-module-library/pull/33

This can be merged even if the rolldown doesn't yet support eth transfers


rollup/node/src/chain_spec.rs
```diff
@@ -294,9 +299,9 @@ fn rollup_genesis(
 			assets: register_assets
 				.iter()
 				.cloned()
-				.map(|(id, meta)| {
+				.map(|(id, meta, maybe_l1_asset)| {
```
majusko, line 302: Why use `maybe` in the name? It's clear from the type that it's optional.

majusko (approved):

rollup/node/src/chain_spec.rs
iStrike7: That is generally how variables of the type `Option` are named atleast in the substrate codebase

majusko: ok, didn't know, I'm used to different approach

PR #735, opened 2024-04-17
Removed eth from genesis and added 2 dummy token asset registry infos

iStrike7:


mateuszaaa (commented):
do we really need dummy token btw?

PR #752, opened 2024-05-13
Remove unused storage item

mateuszaaa:


iStrike7 (changes requested):
This is required for prometheus

PR #783, opened 2024-07-09
Feature/chain genesis salt

iStrike7:


rollup/node/src/chain_spec.rs
```diff
@@ -82,12 +84,21 @@ pub fn rollup_local_config(
 		),
 	};
 
+	let mut chain_genesis_salt_arr = [0u8; 32];
+	if randomize_chain_genesis_salt {
+		thread_rng().fill(&mut chain_genesis_salt_arr[..]);
```
mateuszaaa, line 86: salt will be static per compilation, right ?

iStrike7: no, that is a typical function call it will change each time the node is started unless started with a path to a chainspec file...
I have imporved the impl to have the option to randomize or set the salt from cli... And without any cli flags it will function as usual

mateuszaaa (approved):
lgtm

PR #800, opened 2024-09-05
[MGX-1183] Disable AssetRegistry::register_asset on call filter level

mateuszaaa:


mateuszaaa:
looks like BaseFilter does not apply to sudo calls, so we would have to disable it in orml_tokens repository which seems like too much effort for sudo only call

PR #819, opened 2024-09-26
[MGX-1443] Feature/foundation membership

vanderian:
add membership pallet used to manage foundation members

only allowed method is for the member to change his key


goncer (approved):

iStrike7 (approved):
LGTM

PR #835, opened 2024-11-01
add multiswap_asset as a proxy type...

decentration:
This is so that we can execute xcm calls from sibling chains, specficially Turing chain, which allows for scheduling xcm calls at a later date and/or recurring.

Currently there is only AutoCompound as a ProxyType. But with MultiswapAsset other chains can execute a mangata based swap from their chian.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -1456,6 +1456,7 @@ where
 		)]
 		pub enum ProxyType {
 			AutoCompound,
+			MultiswapAsset
 		}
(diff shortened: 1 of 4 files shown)


chrisli30:
I fully support @decentration’s proposal. The Autocompound proxy type was added last year to enable Turing XCM automation for auto-compounding LP rewards. Having run securely for 12 months without any vulnerabilities, I believe it’s now safe to introduce another proxy type to unlock more automation opportunities.

PR #838, opened 2024-11-15
Feature/stable pool mgx 1307

goncer:


pallets/market/src/benchmarking.rs
vanderian, line 1: any ideas here?

i just did a copypaste of the whole bench and changed xyk -> sswap
we should get the max of the two for the extrinsic

pallets/market/src/lib.rs
```diff
@@ -0,0 +1,1067 @@
...
+			total_supply: T::Balance,
+		},
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn integrity_test() {
+			assert!(true, "template",);
+		}
+	}
```
vanderian, line 256: remove

pallets/market/src/lib.rs
```diff
@@ -0,0 +1,1067 @@
...
+			}
+
+			ensure!(amount_in < max_amount_in, Error::<T>::ExcesiveInputAmount);
+
+			let swaps = Self::do_swaps(&sender, pools, path.clone(), amount_in, asset_amount_out)?;
+
+			Self::deposit_event(Event::AssetsSwapped { who: sender.clone(), swaps });
+
+			// total swaps inc
+
```
vanderian, line 735: todo total swaps storage

pallets/market/src/mock.rs
vanderian, line 1: currently for the purposes of running benchmark tests
might move the tests form runtime-integration herer

pallets/market/src/weights.rs
vanderian, line 1: edit after merge, CI run, copypaste bench values

pallets/proof-of-stake/src/lib.rs
```diff
@@ -1732,6 +1732,10 @@ impl<T: Config> ProofOfStakeRewardsApi<T::AccountId, BalanceOf<T>, CurrencyIdOf<
 	) -> Result<BalanceOf<T>, DispatchError> {
 		Self::calculate_native_rewards_amount(user, liquidity_asset_id)
 	}
+
+	fn rewards_period() -> u32 {
+		Self::rewards_period()
+	}
```
vanderian, line 1738: so we dont have to depened on the whole config in benchmarks thingy to just get some value

pallets/stable-swap/src/lib.rs
```diff
@@ -0,0 +1,1752 @@
...
+				asset_out,
+				amount_out: dy,
+			});
+
+			Ok(SwapResult {
+				amount_out: dy,
+				total_fee: fee,
+				treasury_fee: to_treasury,
+				bnb_fee: to_bnb,
+			})
```
vanderian, line 696: keep for a while for stash easier integration, we should eventually on take the aggregated Market.Swap event

pallets/stable-swap/src/weights.rs
vanderian, line 1: since the stableswap extrinsics wont be exposed, not needed at the moment

pallets/xyk/src/lib.rs
```diff
@@ -3672,3 +3702,102 @@ impl<T: Config> PoolCreateApi<T::AccountId, BalanceOf<T>, CurrencyIdOf<T>> for P
...
+	) -> Option<(Self::Balance, Self::Balance)> {
+		let pool = Self::get_pool_info(pool_id)?;
+		Self::get_burn_amount(pool.0, pool.1, lp_burn_amount).ok()
+	}
+
+	fn get_non_empty_pools() -> Option<Vec<Self::CurrencyId>> {
+		Self::get_liq_tokens_for_trading().ok()
+	}
+
+	fn get_mint_amount(
```
vanderian, line 3749: this is not that fancy
we would need to make a "swap" computation to get the reservers after the swap, and then get the correct values
doable but not worth the effort at the moment

rollup/runtime/integration-test/src/proxy.rs
vanderian, line 1: proxy does not make sense atm, xyk extrinsics removed
keep it just in case we want to build smth later

rollup/runtime/src/runtime_config.rs
```diff
@@ -1040,10 +837,12 @@ pub mod config {
...
 					// merge the imbalance caused by paying the fees and refunding parts of it again.
-					let adjusted_paid = paid
-						.offset(refund_imbalance)
-						.same()
-						.map_err(|_| TransactionValidityError::Invalid(InvalidTransaction::Payment))?;
+					let adjusted_paid = match paid.offset(refund_imbalance) {
+						SameOrOther::Same(a) => Ok(a),
+						SameOrOther::None => Ok(C::NegativeImbalance::from_zero(token_id)),
+						SameOrOther::Other(b) => Err(b),
+					}
```
vanderian, line 844: `SameOrOther::same()` in case of None used the `Default::default()`
added a PR to orml to remove the `Default` impl for imbalanced and replace with `fail!(...)`

rollup/runtime/src/runtime_config.rs
```diff
@@ -999,15 +795,16 @@ pub mod config {
 					ExistenceRequirement::KeepAlive,
 				) {
 					Ok(imbalance) => Ok(Some(LiquidityInfoEnum::Imbalance((T1::get(), imbalance)))),
-					// TODO make sure atleast 1 planck KSM is charged
+					Err(_) if fee_2.is_zero() => Err(InvalidTransaction::Payment.into()),
```
vanderian, line 798: it can happen that a fee would be zero in the second asset
fail rather then withdraw 1 token

rollup/runtime/src/runtime_config.rs
```diff
@@ -1285,6 +1091,81 @@ pub mod config {
...
+			type StringLimit = T::StringLimit;
+
+			fn metadata(
+				asset_id: &Self::AssetId,
+			) -> Option<AssetMetadata<Self::Balance, Self::CustomMetadata, Self::StringLimit>> {
+				orml_asset_registry::Pallet::<T>::metadata(asset_id)
+			}
+		}
+
+		fn format_u128_with_leading_zeros(num: TokenId, width: usize) -> Vec<u8> {
```
vanderian, line 1147: replacement for the thingy we have XYK to get the symbols...a bit more readable version

rollup/runtime/src/runtime_config.rs
```diff
@@ -1285,6 +1091,81 @@ pub mod config {
 			fn get_asset_l1_id(asset_id: T::AssetId) -> Option<L1Asset> {
 				orml_asset_registry::IdToL1Asset::<T>::get(asset_id)
 			}
+
+			fn create_pool_asset(
```
vanderian, line 1095: the asset mutation trait is in XYK pallet, didn't want to touch XYK so added a fn to this trait

rollup/runtime/src/runtime_config.rs
```diff
@@ -1336,12 +1217,7 @@ pub mod config {
 				let call: crate::CallType = (c.clone()).into();
 
 				match call {
-					CallType::MultiSell { .. } |
-					CallType::MultiBuy { .. } |
-					CallType::AtomicBuy { .. } |
-					CallType::AtomicSell { .. } |
-					CallType::CompoundRewards |
-					CallType::ProvideLiquidityWithConversion => true,
```
vanderian, line 1344: xyk calls are removed, so all of the stuff needs to go...

scripts/build-image.sh
```diff
@@ -15,7 +15,7 @@ else
     DOCKER_LABEL=${GIT_REV}-dirty
 fi
 
-DOCKER_IMAGE_TAG=${@:-mangatasolutions/rollup-node:dev}
+DOCKER_IMAGE_TAG=${@:-mangatasolutions/rollup-node:local}
```
vanderian, line 18: the local scripts and docs provide build instr with `local` tag not `dev`

pallets/market/src/lib.rs
```diff
@@ -0,0 +1,1092 @@
...
+						first_asset_id,
+						first_asset_amount,
+						second_asset_id,
+						second_asset_amount,
+					)?;
+					lp_token
+				},
+				PoolKind::StableSwap => {
+					let first_decimal = Self::get_decimals(&first_asset_id)?;
+					let second_decimal = Self::get_decimals(&second_asset_id)?;
```
goncer, line 297: Why do we care about decimals? do not fully understand this,

pallets/market/src/lib.rs
```diff
@@ -0,0 +1,1092 @@
...
+
+					let lp_token = T::StableSwap::create_pool(
```
goncer, line 299: do not we need the amplification parameter somewhere? :thinking:

pallets/market/src/tests.rs
```diff
@@ -0,0 +1 @@
+
```
goncer, line 1: ups?

vanderian: it is a feature so all the tests pass :D

pallets/market/src/lib.rs
vanderian: you can have two tokens
USDT with 6 decimals
DAI with 18 decimals
1_000_000 == 1_000_000_000_000_000_000 should be equal, so we need to take that into account somehow

goncer: but,.. should we care about the rate? 
I mean, I guess that I should be able to create the following, 
TokenA: Dollar
TokenB: 50cCoin representation

I guess that I should be able to create a stable_pool with ( 1 USD, 2 cents ) , Or are we restricting the liquidity we put on both sides somehow?

PR #841, opened 2024-11-22
GASP-1625 Feature/locked tokens

vanderian:
adds checks for non transferable tokens

`--features=unlocked` removes such restrictions

runtime cofing defines the native currency - `0` as non transferable token

tokens all transfer calls, set_balance, mint extrinsics are checked
market create, mint, burn are checked
market swaps are allowed

foundation members have veto over:
market create_pool & burn


rollup/runtime/src/runtime_config.rs
```diff
@@ -48,6 +48,11 @@ pub mod tokens {
 	pub const RX_TOKEN_ID: TokenId = 0;
 	pub const ETH_TOKEN_ID: TokenId = 1;
 
+	#[cfg(any(feature = "unlocked", feature = "runtime-benchmarks"))]
+	pub type NontransferableTokens = Nothing;
+	#[cfg(not(feature = "unlocked"))]
+	pub type NontransferableTokens = Equals<ConstU32<RX_TOKEN_ID>>;
+
```
vanderian: all allowed for eg testnet
otherwise just native token '0'

pallets/market/src/lib.rs
```diff
@@ -567,6 +607,13 @@ pub mod pallet {
...
 
 			let pool_info = Self::get_pool_info(pool_id)?;
+			// check assets id, or the foundation has a veto
+			ensure!(
+				(!T::NontransferableTokens::contains(&pool_info.pool.0) &&
+					!T::NontransferableTokens::contains(&pool_info.pool.1)) ||
+					T::FoundationAccountsProvider::get().contains(&sender),
+				Error::<T>::NontransferableToken
+			);
 			Self::check_assets_allowed(pool_info.pool)?;
```
goncer, line 641: I think there must be no restriction about burning

pallets/market/src/lib.rs
```diff
@@ -377,6 +399,12 @@ pub mod pallet {
 			let sender = ensure_signed(origin)?;
 
 			let pool_info = Self::get_pool_info(pool_id)?;
+			// check assets id, foundation has no veto
+			ensure!(
+				!T::NontransferableTokens::contains(&pool_info.pool.0) &&
+					!T::NontransferableTokens::contains(&pool_info.pool.1),
+				Error::<T>::NontransferableToken
+			);
 			Self::check_assets_allowed(pool_info.pool)?;
```
goncer, line 423: perhaps the foundation must be able to do this. how do they will mint more liq in the case the pool go bananas?

goncer (approved):

pallets/market/src/lib.rs
vanderian: a pool with GASP can be burned only by foundation account
other pools are allowed
the way it is in the specs table

technically this case should not happen, since onely foundation can create pool -> therefore has all the liq tokens -> is the only account that can burn them

pallets/market/src/lib.rs
vanderian: agree, but the specs say otherwise
we can always sell with arbitrage bot to level the pool

goncer (approved):

PR #850, opened 2024-12-02
CI: Add additionoal unlocked runtime image build in CI

tenequm:
Introduce a separate build process for Docker images with an unlocked runtime, along with the necessary adjustments to export WASM artifacts and manage image manifests.


goncer (approved):
:crossed_fingers:

PR #877, opened 2025-01-13
[GASP-1826] reverts swap trx fees

vanderian:
swaps always pays commission
min fee fix
fee goes to trsy, bnb, pool
fix valuation where it applies to the real, not predicted swap value


pallets/fee-lock/src/benchmarking.rs
```diff
@@ -105,7 +105,7 @@ benchmarks! {
 		#[cfg(not(test))]
 		assert_ok!(<T as Config>::Xyk::create_pool(caller, valuating_token_id, pool_amount, token_id, pool_amount.saturating_mul(2u8.into())));
 
-		let value: BalanceOf<T> = 1980_u32.into();
+		let value: BalanceOf<T> = 2000_u32.into();
```
mateuszaaa, line 108: does it meat that some rounding has change? that will explode in e2e i guess ..

pallets/stable-swap/src/lib.rs
```diff
@@ -839,6 +839,23 @@ pub mod pallet {
 			Ok(BoundedVec::truncate_from(amounts))
 		}
 
+		pub fn settle_pool_fees(
```
mateuszaaa, line 842: is it used anywhere in runtime ?

pallets/fee-lock/src/benchmarking.rs
vanderian: valuation changed to current state formula vs expected swap formula, so the test value had to change too

previously we had the valuation in fee lock pallet, then it was cleaned up and used a trait from market pallet that relays further to XYK
but the XYK method was using calculate_sell_price_no_fee which takes into account the expected swap amount
there is a fix to replace that method with actual state value as stated in docs 
check change in XYK pallet

yes, unfortunately e2e will have to adapt, but only for cases related to valuation

pallets/stable-swap/src/lib.rs
vanderian: yes, down below in correct_and_deposit
```
Some(LiquidityInfoEnum::FeeLockSwap((pool, fee_asset, fee))) => {
	if post_info.pays_fee == Pays::Yes {
		FeeHelpers::<T, C, OFLA>::settle_fees(who, pool, fee_asset, fee)
```

mateuszaaa (approved):
lgtm

